PR #74, opened 2022-07-13
Task | ENG-2573 | Diego Maldonado

DiegoMGonet:
* Send hardware reading


NeuroID/src/androidLib/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -55,14 +55,27 @@ class NeuroID private constructor(
...
         application?.let {
             NIDSharedPrefsDefaults(it).setUserId(userId)
         }
         getDataStoreInstance().saveEvent(
             NIDEventModel(
                 type = SET_USER_ID,
                 uid = userId,
-                ts = System.currentTimeMillis()
+                ts = System.currentTimeMillis(),
+                gyro = NIDSensorModel(
```
alejandrobautistar, line 69: We can return NIDSensorModel on NIDSensorHelper.get Gyroscope Info() and avoid all that instances created

PR #147, opened 2023-05-08
ENG-5004 - Add Integration Health Functions

kevinsites:


version.properties
cometman, line 1: This should stay as version of the SDK. We can version this one as 2.7.0

cometman (changes requested):

PR #148, opened 2023-06-21
v2.7.1

cometman:


version.properties
```diff
@@ -1,5 +1,5 @@
 #Fri May 27 05:08:14 UTC 2022
-SDK_VERSION=2.6.4
-VERSION_MAJOR=2
-VERSION_MINOR=6
-VERSION_PATCH=4
+SDK_VERSION=Integration report
+VERSION_MAJOR=Integration report
+VERSION_MINOR=Integration report
+VERSION_PATCH=Integration report
```
kevinsites, line 5: I don't think this is right is it?

kevinsites (changes requested):

version.properties
cometman: LOL we gotta update that logic. I forget its automated based on PR name

PR #163, opened 2023-08-03
ENG-5877 add cast check and return not_a_view message.

kishisaka-nid:
Added cast check prior to doing the cast. If not a view, we will send back a small message “not_a_view” which will be added to the screen_hierarchy section of the event during the reporting. Code below invokes the getParentsOfView() in getParents().

```
val classJson = JSONObject().put("n", "screenHierarchy")
    .put("v", "${view.getParents()}${NIDServiceTracker.screenName}")
```

If this is not how we should handle non views during the layout traversal then please comment on how this should be handled (Throw Error or a differnt message perhaps). Thanks.

null or is not View cases will result in screenHierarchy values being sent to the server that looks like this: 
`{"n":"screenHierarchy","v":"not_a_viewPERSONAL_DETAILS"},`


cometman (changes requested):
Can you also make this change in the React native lib?

https://github.com/Neuro-ID/neuroid-android-sdk/blob/master/NeuroID/src/reactNativeLib/java/com/neuroid/tracker/utils/NIDViewsExtensions.kt#L33

NeuroID/src/androidLib/java/com/neuroid/tracker/utils/NIDViewsExtensions.kt
```diff
@@ -39,9 +39,14 @@ private fun getParentsOfView(layers: Int, view: View): String {
     if (view.parent == null) {
         throw Error("Before registerTarget add the view to his parent")
     }
-    val childView = view.parent as View
-    return if (layers == 3 || childView.id == android.R.id.content) "" else {
-        "${childView.javaClass.simpleName}/${getParentsOfView(layers + 1, childView)}"
+    return if (view.parent is View) {
+        valew = view.parent as View
```
kevinsites, line 43: 2 q's, is it `valew` or `childView`? and do we need this? I don't see it used anywhere?

kevinsites (changes requested):

NeuroID/src/androidLib/java/com/neuroid/tracker/utils/NIDViewsExtensions.kt
kishisaka-nid: yikes, thanks for that, incorrect syntax there. should be fixed now.

kishisaka-nid:
will add that to the react native one as well.

kevinsites (approved):

NeuroID/src/reactNativeLib/java/com/neuroid/tracker/utils/NIDViewsExtensions.kt
```diff
@@ -31,9 +31,17 @@ fun View.getParents(): String {
 }
 
 private fun getParentsOfView(layers: Int, view: View): String {
-    val childView = view.parent as View
-    return if (layers == 3 || childView.id == android.R.id.content) "" else {
-        "${childView.javaClass.simpleName}/${getParentsOfView(layers + 1, childView)}"
+    if (view.parent == null) {
+        throw Error("Before registerTarget add the view to his parent")
```
cometman, line 35: What is the purpose of this error?

kishisaka-nid: good point, removed the check so that we don't throw potentially app crashing exceptions that aren't handled. the cast check will handle the null case as well and in this case we will send back "not_a_view" Thanks.

kevinsites (approved):

cometman (approved):

PR #167, opened 2023-08-22
ENG-5952 duplicate event edittext fix and text watcher cut/paste  fix 

kishisaka-nid:
Fix for the condition where edit text fields will add duplicate text watchers when app is backgrounded then foregrounded. This will send unwanted additional text input events back the server. This is fixed in the release/debug/reactnative flavors. 

Fix for the condition where a registered text watcher will send an unwanted paste event on every text input after a paste into the edit text field is done. All flavor fix.


kishisaka-nid:
broken when switching between activities.

PR #183, opened 2023-08-29
ENG-5667-updated build gradle and corrected project path

priya-xavier:
Made changes such that the aar files gets published to Jitpack.io!


kishisaka-nid (approved):
pulled, tested with command "./gradlew clean -Pgroup=com.github.kishisaka-nid -Pversion=v3.0.8 -xtest -xlint assemble publishToMavenLocal"
 LGTM 🚢 it!

NeuroID/build.gradle
```diff
@@ -89,7 +89,7 @@ dependencies {
 
     implementation("androidx.lifecycle:lifecycle-process:2.4.1")
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-android:1.6.1")
-    reactNativeLibImplementation("com.facebook.react:react-native:+")
+    reactNativeLibImplementation("com.facebook.react:react-native:0.71.0-rc.0")
```
kishisaka-nid, line 92: Good one!

build.gradle
```diff
@@ -2,8 +2,8 @@
 buildscript {
     repositories {
         google()
-        jcenter()
         mavenCentral()
+        jcenter()
```
kevinsites, line 6: Just curious, does the order here matter?

kevinsites (approved):

quacke-nid (approved):

build.gradle
priya-xavier: It will pull libraries from those repositories in that order. jcenter is deprecated to read only so we should eventually move away from pulling react-native from jcenter. Can't do without it just yet since though.

PR #193, opened 2023-09-11
ENG-6122 have setScreenName() throw exception is SDK is not started

kishisaka-nid:
Have setScreenName() throw exception is SDK is not started


kevinsites (approved):
Do we need to add this in the RN flavor?

kishisaka-nid:
no NeuroID class in the RN flavor to update AFAIK.

priya-xavier (approved):

PR #199, opened 2023-09-20
ENG-6208 - Observe changes in activity children

kevinsites:
The fix for RN is to uncomment out the line we removed in the Android Native SDK because it was causing double event registration, however because of how RN creates/destroys activities we have to put this back in to capture all the changes when a user goes to a new page. If you have more questions lmk

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/utils/NIDTextWatcher.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/utils/NIDTextWatcher.kt
@@ -19,24 +19,18 @@ import org.json.JSONObject
 
 class NIDTextWatcher(
     private val idName: String,
-    val className: String? = ""
+    val className: String? = "",
+    val startingHashValue: String? = ""
 ) : TextWatcher {
 
     private var lastSize = 0
+    private var lastHashValue = startingHashValue
 
     override fun beforeTextChanged(p0: CharSequence?, p1: Int, p2: Int, p3: Int) {
         // No operation
     }
 
     override fun onTextChanged(sequence: CharSequence?, start: Int, before: Int, count: Int) {
-        /**
-         * Potentially check for paste here
-         */
-//        NIDLog.d(
-//            "NIDDebugEvent",
-//            "**OnTextChange $idName - $className - $sequence - $start - $before - $count"
-//        )
-
 
         // Check if the change is due to a paste operation
         val clipboard = NeuroID.getInstance()?.getApplicationContext()
@@ -48,7 +42,7 @@ class NIDTextWatcher(
                 pastedText = clipData.getItemAt(0).text.toString()
             } catch (e: Exception) {
                 e.message?.let {
-                    NIDLog.e("NID-Activity",it)
+                    NIDLog.e("NID-Activity", it)
                 }
             }
             val pasteCount = pastedText.length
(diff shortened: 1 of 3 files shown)


NeuroID/src/reactNativeLib/java/com/neuroid/tracker/callbacks/NIDActivityCallbacks.kt
```diff
@@ -124,6 +124,16 @@ class NIDActivityCallbacks() : ActivityLifecycleCallbacks {
...
+
+
+        val currentActivityName = activity::class.java.name
+        registerTargetFromScreen(
+            activity,
+            registerTarget = true,
+            registerListeners = true,
+            activityOrFragment = "activity",
+            parent = currentActivityName
+        )
```
kevinsites, line 136: @kishisaka-nid  - Unsure if we need this in my PR or if your work will cover it, but left it in because of our session yesterday

kishisaka-nid: Yeah we need this otherwise the registrations don't occur and we don't get edited events.

priya-xavier (approved):

NeuroID/src/main/java/com/neuroid/tracker/utils/NIDTextWatcher.kt
```diff
@@ -88,25 +82,29 @@ class NIDTextWatcher(
...
+                            "attr" to getAttrJson(sequence.toString()),
+                            "etn" to INPUT,
+                            "et" to "text"
+                        ),
+                        tgs = idName,
+                        v = "S~C~~${sequence?.length}",
+                        hv = sequence?.toString()?.getSHA256withSalt()?.take(8),
+                        gyro = gyroData,
+                        accel = accelData
+                    )
```
kevinsites, line 106: @kishisaka-nid  - Here is the new checking mechanism

NeuroID/src/reactNativeLib/java/com/neuroid/tracker/callbacks/NIDActivityCallbacks.kt
kevinsites: @kishisaka-nid  - I removed this because we are now going to be calling forceStart on every page and if the user does go in and out of the app I didn't want to re-register things and maybe have another duplicate listener issue

NeuroID/src/reactNativeLib/java/com/neuroid/tracker/events/NIDIdentifyAllViews.kt
```diff
@@ -287,7 +287,11 @@ private fun registerListeners(view: View) {
             "EditText Listener $simpleClassName - ${view::class} - ${view.getIdOrTag()}"
         )
         // add Text Change watcher
-        val textWatcher = NIDTextWatcher(idName, simpleClassName)
+        val textWatcher = NIDTextWatcher(
+            idName,
+            simpleClassName,
+            "${view.text.toString().getSHA256withSalt()?.take(8)}"
+        )
```
kishisaka-nid, line 294: This will take care of the empty starting case right?

NeuroID/src/main/java/com/neuroid/tracker/utils/NIDTextWatcher.kt
```diff
@@ -19,24 +19,18 @@ import org.json.JSONObject
...
 
     override fun onTextChanged(sequence: CharSequence?, start: Int, before: Int, count: Int) {
-        /**
-         * Potentially check for paste here
-         */
-//        NIDLog.d(
-//            "NIDDebugEvent",
-//            "**OnTextChange $idName - $className - $sequence - $start - $before - $count"
-//        )
-
```
kishisaka-nid, line 39: suggest deleting this commented section

kishisaka-nid (approved):
Pulled, tested, filters out duplicate events as advertised. 🚢 it!

NeuroID/src/main/java/com/neuroid/tracker/utils/NIDTextWatcher.kt
```diff
@@ -88,25 +82,29 @@ class NIDTextWatcher(
         val ts = System.currentTimeMillis()
         val gyroData = NIDSensorHelper.getGyroscopeInfo()
         val accelData = NIDSensorHelper.getAccelerometerInfo()
+        val currentHashValue = sequence?.toString()?.getSHA256withSalt()?.take(8)
+        
+        if (lastHashValue != currentHashValue) {
```
cometman, line 87: Will this prevent an input if a user pastes in a value of the same value?

PR #202, opened 2023-09-21
ENG-6220 - Add Support for RN Buttons

kevinsites:
Because of how RN renders buttons (a view with either a text or image inside) we needed to specifically look for them.

NOTE: This will not capture the click/touch events that happen on said buttons as Android does not allow more than 1 click listener and RN is using that spot. But we can at least register the target so that DS can see what the page looks like

NOTE: This PR will capture basic included RN elements, however because you can install any 3rd party library for other RN components I don't know that its worth the effort to try and find every possible element and include it, especially as TextFields seem to provide the most value for DS


kishisaka-nid (approved):
pulled, tested, I am seeing react native images in the react view group in the event upload `"eid":"ReactViewGroup_660_00-ReactImageView_00_530"`, 
🚢

priya-xavier (approved):
Confirming that I'm also seeing the Register target event for the text view RN Button on the first page. I don't see this for the buttons on the rest of the pages (assuming that's the note in the PR is about).

PR #204, opened 2023-09-26
ENG-6239 - Add isRN var and methods to use it

kevinsites:
A lot of noise in this PR because I moved the `createMobileMetadata` event down to be close to where it is being called. The only real changes are:
- Adding `isRN` bool var to the NeuroID Class
- Adding `setIsRN` method to mark `isRN` as true
- Using isRN in the `getSDKVersion` method
- Appending a `isRN` key/value to the mobile metadata event

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/utils/NIDVersion.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/utils/NIDVersion.kt
@@ -1,10 +1,19 @@
 package com.neuroid.tracker.utils
 
 import com.neuroid.tracker.BuildConfig
+import com.neuroid.tracker.NeuroID
 
 object NIDVersion {
     fun getSDKVersion(): String {
-        return "5.android-" + BuildConfig.VERSION_NAME
+        val rnText = NeuroID.getInstance()?.isRN?.let {
+            if (it) {
+                "rn-"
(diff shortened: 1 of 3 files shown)


quacke-nid (approved):

NeuroID/src/main/java/com/neuroid/tracker/utils/NIDVersion.kt
```diff
@@ -1,10 +1,19 @@
 package com.neuroid.tracker.utils
 
 import com.neuroid.tracker.BuildConfig
+import com.neuroid.tracker.NeuroID
 
 object NIDVersion {
     fun getSDKVersion(): String {
-        return "5.android-" + BuildConfig.VERSION_NAME
+        val rnText = NeuroID.getInstance()?.isRN?.let {
```
kishisaka-nid, line 8: consider: 

var rnText = ""
rnText = NeuroID.getInstance()?.isRN?.let {....

just in case NeuroID.getInstance() returns null

kevinsites: I tried to put that and it failed because the `NeuroID.getInstance()?.isRN?.let` returns an optional string and `var rnText` is a non-optional string

PR #206, opened 2023-09-28
ENG-6236-adding check for duplicate paste events

priya-xavier:


kishisaka-nid (approved):
pulled, tested, saw only one paste event, address the other comments and 🚢 

```
2023-09-28 14:31:25.498   875-875   NIDDebug                com...idreactnativesdksandbox.debug  D  Long Press Action Context - Paste
2023-09-28 14:31:25.500   875-964   NIDDebugEvent           com...idreactnativesdksandbox.debug  D  EVENT: CONTEXT_MENU - null - meta=null
2023-09-28 14:31:25.503   875-964   NIDDebugEvent           com...idreactnativesdksandbox.debug  D  EVENT: PASTE - First Name - 
2023-09-28 14:31:25.522   875-875   NID-Activity            com...idreactnativesdksandbox.debug  D  after text fsdffsdfsdfsdf
2023-09-28 14:31:25.523   875-964   NIDDebugEvent           com...idreactnativesdksandbox.debug  D  EVENT: INPUT - First Name - v=S~
```

NeuroID/src/main/java/com/neuroid/tracker/utils/NIDTextWatcher.kt
```diff
@@ -20,11 +20,13 @@ import org.json.JSONObject
 class NIDTextWatcher(
     private val idName: String,
     val className: String? = "",
-    val startingHashValue: String? = ""
+    val startingHashValue: String? = "",
+    val startingPastedHashValue: String? = ""
```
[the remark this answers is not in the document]
priya-xavier: removed the unused start var.

NeuroID/src/main/java/com/neuroid/tracker/utils/NIDTextWatcher.kt
```diff
@@ -48,32 +50,38 @@ class NIDTextWatcher(
             val pasteCount = pastedText.length
             if (sequence.toString().contains(pastedText) && (pasteCount == count)) {
                 // The change is likely due to a paste operation
-                val ts = System.currentTimeMillis()
-                val gyroData = NIDSensorHelper.getGyroscopeInfo()
-                val accelData = NIDSensorHelper.getAccelerometerInfo()
 
-                val metadataObj = JSONObject()
-                metadataObj.put("clipboardText", "S~C~~${pastedText.length}")
+                val currentPastedHashValue = sequence?.toString()?.getSHA256withSalt()?.take(8)
```
kishisaka-nid, line 53: Might not need to use the SHA256. String has a hashCode() that should suffice for this. Can consider updating next time.

priya-xavier: updated to use hashCode. Thanks!

kevinsites (approved):
LGTM, the only thing to note which we should run by DS is that if I copy the field value and paste it to replace the value then it will not show a paste event (I.e. value = "Priya", I copy that and then select the whole value and paste "Priya" it will show an input but not paste). I don't know if its a big deal because it would be the same value and nothing would have changed but something to note

PR #221, opened 2023-11-06
ENG-6512 & ENG-6557 & ENG-6510 - MIHR dependency, Throw on Start, Simplify Configure

kevinsites:


NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -263,7 +306,16 @@ class NeuroID private constructor(
...
     }
 
-   open fun start() {
+    open fun start() {
+        if (clientKey == "") {
+            NIDLog.e(
+                "NeuroID Error",
+                "Missing Client Key - please call configure prior to calling start"
+            )
+            throw IllegalArgumentException("NeuroID SDK Missing Client API Key");
```
quacke-nid, line 315: Is there a more appropriate exception?  There are no arguments to `start`.

quacke-nid (approved):

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -93,6 +105,17 @@ class NeuroID private constructor(
         fun getInstance(): NeuroID? = singleton
     }
 
+    internal fun validateClientKey(clientKey: String): Boolean {
```
kishisaka-nid, line 108: Might wanna add a test for this one.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: maybe IllegalStateException might be better.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -142,21 +165,26 @@ class NeuroID private constructor(
...
+        )
     }
 
     fun getEnvironment(): String = NIDServiceTracker.environment
 
     fun setSiteId(siteId: String) {
+        NIDLog.i(
+            "NeuroID Info",
+            "**** NOTE: setSiteId METHOD IS DEPRECATED"
+        )
```
priya-xavier, line 187: Should we also have this log for the getSiteId() method since we won't have the site id set by the user going forward?

kevinsites: We want a message that its deprecated but we aren't removing it completely yet so that when people upgrade it doesn't break

priya-xavier: I was asking about the _get_ siteId method. We are not using that method in the sdk as far as I can see. So does that also need to be deprecated?

kevinsites: Oh I misread, good catch. Will add a message now

priya-xavier (dismissed):

kishisaka-nid (dismissed):
pulled, tested, saw exceptions when invalid stuff was put in, no exceptions when all is correct. 🚢 it!

priya-xavier (approved):

kishisaka-nid (approved):

PR #224, opened 2023-11-08
Eng 5590 simplify responseID generation. 

kishisaka-nid:
Removed the MD5 hasher and the Calendar code that generates the response ID. The response ID in the JSON request body will now be a UUID that will look like this:  
"responseId":"8823db97-8d46-43c4-be5d-624dd7f0726b"

Currently in master (Android), the JSON request body will return response IDs like this: 
"responseId":"f25588839fffdc00"

Checked with Umut and Jeff. Jeff mentioned the backend can handle this response ID format and Umut mentioned ENR doesn't look at this parameter. We should be good.


priya-xavier (approved):
LGTM!

kevinsites (approved):

PR #225, opened 2023-11-08
ENG-6571-added FPJS endpoint environment url

priya-xavier:
Currently only the PROD endpoint is going to be available in the sdk. The dev endpoint is only for internal testing.

### DIFF
--- a/NeuroID/src/advancedDeviceAndroidLib/java/com/neuroid/tracker/extensions/AdvancedDeviceExtension.kt
+++ b/NeuroID/src/advancedDeviceAndroidLib/java/com/neuroid/tracker/extensions/AdvancedDeviceExtension.kt
@@ -16,6 +16,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import com.neuroid.tracker.utils.FPJSHelper
 import com.neuroid.tracker.utils.NIDLogWrapper
+import com.neuroid.tracker.utils.Constants
 
 fun NeuroID.start(advancedDeviceSignals: Boolean) {
     start()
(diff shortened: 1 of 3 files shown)


NeuroID/src/advancedDeviceAndroidLib/java/com/neuroid/tracker/extensions/AdvancedDeviceExtension.kt
```diff
@@ -25,13 +26,18 @@ fun NeuroID.start(advancedDeviceSignals: Boolean) {
             val keyService = NIDAdvKeyService()
             var fpjsRetryCount = 0
             val FPJS_RETRY_MAX = 3
+            var endpointUrl = Constants.fpjsDevDomain.displayName
+            if (NeuroID.getInstance()?.getEnvironment() == "LIVE"){
+                endpointUrl = Constants.fpjsProdDomain.displayName
+            }
```
kevinsites, line 32: This is my fault for putting it in the ticket (I was making it quick). Although there is a dev option, we should not be putting that in the SDK as our customers regardless of their `env` will only ever call PROD. Only we as internal NID devs will call the dev endpoint.

kevinsites (changes requested):

NeuroID/src/advancedDeviceAndroidLib/java/com/neuroid/tracker/extensions/AdvancedDeviceExtension.kt
priya-xavier: Ok. I removed the dev endpoint completely from the sdk.

priya-xavier: Should we document the dev endpoint in the readme?

priya-xavier: Adding the comment in the Jira ticket for documentation.

kevinsites (approved):

kishisaka-nid (approved):

NeuroID/src/advancedDeviceAndroidLib/java/com/neuroid/tracker/extensions/AdvancedDeviceExtension.kt
priya-xavier: Tested with site id and key for adv enabled form_1oLmFMjl and its working as expected.

PR #231, opened 2023-11-21
ENG-6705 Update resource ID call in app

priya-xavier:
Update call in the sample app to match sdk function `excludeViewByTestID`


kishisaka-nid (approved):
Only happens in the release branch. LGTM. 🚢 it!

asnodgrass-nid (approved):

PR #234, opened 2023-11-28
ENG-6176 text watcher tests

priya-xavier:


priya-xavier:
Updating tests via a different way of mock injection in a different PR.

PR #236, opened 2023-11-29
ENG-6728 - Add new session commands

kevinsites:


NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -519,4 +519,82 @@ class NeuroID private constructor(
...
 
     fun getSDKVersion() = NIDVersion.getSDKVersion()
+
+    // new Session Commands
+    fun clearSessionVariables() {
+        userID = ""
+//        registeredUserID = ""
+    }
+
+    fun startSession(sessionID: String = ""): Pair<Boolean, String> {
```
kishisaka-nid, line 529: Instead of returning a pair, returning a class instance is better like a SessionResult `data class SessionResult(var result: Boolean, var sessionId: String)` or something like that.

kishisaka-nid (dismissed):
LGTM! 🚢 it!

priya-xavier (dismissed):

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -519,4 +520,82 @@ class NeuroID private constructor(
...
+        if (clientKey == "") {
+            NIDLog.e(
+                msg = "Missing Client Key - please call configure prior to calling start"
+            )
+            return SessionStartResult(false, "")
+        }
+
+        if (userID != "" || isSDKStarted) {
+            stopSession()
+        }
```
priya-xavier, line 540: Tiny question, why would we stop the session if the sdk is started? Is it to prevent multiple sessions in progress?

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -519,4 +520,82 @@ class NeuroID private constructor(
...
+            sessionID
+        } else {
+            generateUniqueHexId()
+        }
+
+        if (!setUserID(finalSessionID)) {
+            return SessionStartResult(false, "")
+        }
+
+        isSDKStarted = true
```
priya-xavier, line 552: Is this redundant since it is also set to true in the resumeCollection call?

kevinsites: it is semi redundant but we needed the isSDKStarted variable to be true before createSession so it captures the events

priya-xavier (approved):

cometman (approved):

kishisaka-nid (approved):

PR #239, opened 2023-12-05
ENG-6730 call in progress spike

priya-xavier:
Not to be merged- Listener added to example app to log state changes in phone call status.


kevinsites:
Work is captured in https://github.com/Neuro-ID/neuroid-android-sdk/pull/251

PR #240, opened 2023-12-15
ENG-6167 add tests for the NIDJobServiceManager

kishisaka-nid:
Add tests for the NIDJobServiceManager class. Refactor NIDJobServiceManager to allow mocking and to read internal variables to assert.

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
@@ -2,6 +2,8 @@ package com.neuroid.tracker.service
 
 import android.app.Application
 import com.neuroid.tracker.callbacks.NIDSensorHelper
+import com.neuroid.tracker.storage.NIDDataStoreManager
+import com.neuroid.tracker.storage.getDataStoreInstance
 import com.neuroid.tracker.utils.NIDLogWrapper
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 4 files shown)


NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
```diff
@@ -16,14 +18,14 @@ import java.util.concurrent.TimeUnit
...
+    internal var jobCaptureEvents: Job? = null
     var isSendEventsNowEnabled = true
 
     @Volatile
     var userActive = true
-    private var clientKey = ""
+    internal var clientKey = ""
     private var application: Application? = null
-    private var endpoint: String = ""
+    internal var endpoint: String = ""
```
cometman, line 28: Are we still supporting configuring endpoints @kevinsites ?

kevinsites: Naw we can probably remove it, I am pretty sure we have removed all the references to it anyways

kevinsites (approved):

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
kishisaka-nid: I think we should keep this in there. We do have a development endpoint as well (https://receiver.neuro-dev.com) that we support.

priya-xavier (approved):

PR #242, opened 2024-01-08
ENG-6920 use new session commands fix

kishisaka-nid:
This fix is for a crash that was found during sample app testing. 

If user calls startSession("bad session id") with a bad session id, then calls stopSession(), app will crash because the job manager was never initialized with the proper endpoint (empty) and will try to flush the empty queue causing retrofit to throw an invalid URL exception. To fix this, we statically set the URL to always point to production. This way it will always be set and the call to stopSession() will harmlessly flush an empty queue. 

Also the NeuroID.userID member variable was not being updated when the startSession("") was invoked with an empty id parameter causing getUserID() to return empty. This should have been updated with a random id. The user id was however being updated in the shared prefs and being sent back the server via server events. This was missed. This is fixed now. Tests were added to insure that any future changes made here will ensure that this still works.

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
@@ -98,7 +98,6 @@ class NeuroID private constructor(
     companion object {
         var showLogs: Boolean = true
         var isSDKStarted = false
-        const val ENDPOINT_PRODUCTION = "https://receiver.neuroid.cloud/"
 
         private var singleton: NeuroID? = null
 
--- a/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
@@ -37,7 +37,6 @@ class NeuroID private constructor(
     internal var application: Application?, internal var clientKey: String
 ) {
     private var firstTime = true
-    private var endpoint = ENDPOINT_PRODUCTION
     internal var sessionID = ""
     internal var clientID = ""
     internal var userID = ""
@@ -427,7 +426,7 @@ class NeuroID private constructor(
             saveIntegrationHealthEvents()
         }
         application?.let {
-            nidJobServiceManager.startJob(it, clientKey, endpoint)
+            nidJobServiceManager.startJob(it, clientKey)
         }
         dataStore.saveAndClearAllQueuedEvents()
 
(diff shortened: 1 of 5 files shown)


priya-xavier (approved):
Tested on sample app. LGTM!

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -37,7 +37,6 @@ class NeuroID private constructor(
     internal var application: Application?, internal var clientKey: String
 ) {
     private var firstTime = true
-    private var endpoint = ENDPOINT_PRODUCTION
```
kevinsites, line 40: Where is ENDPOINT_PRODUCTION coming from? Should we remove it where its declared or replace it with the new constant?

NeuroID/src/test/java/com/neuroid/tracker/NIDJobServiceManagerTest.kt
```diff
@@ -31,7 +31,7 @@ class NIDJobServiceManagerTest {
         injectMockedApplication()
         assertEquals(NIDJobServiceManager.isSetup, true)
         assertEquals(NIDJobServiceManager.clientKey, "clientKey")
-        assertEquals(NIDJobServiceManager.endpoint, "endpoint")
+        assertEquals(NIDJobServiceManager.endpoint, "https://test.com")
```
kevinsites, line 34: shouldn't this fail because it should use the constant?

kevinsites (changes requested):

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: Yeah will remove ENDPOINT_PRODUCTION. It has been replaced with Constants.productionEndpoint.displayName. Nice find.

NeuroID/src/test/java/com/neuroid/tracker/NIDJobServiceManagerTest.kt
kishisaka-nid: Had to fix testGetApiService() as well since this is the one that was setting it. Running testStartJob() by itself had no issue but since testGetApiService() was setting this to http://test.com, this was making it pass. Once testGetApiService() was patched with proper endpoint, testStartJob() started to fail. This was also patched so we are good now.

kevinsites (approved):

kevinsites (approved):

PR #244, opened 2024-01-11
ENG-6952, fix setUserId/setRegisteredUsedID

kishisaka-nid:
ensure that setUserID () and setRegisteredUserID() returns false and not set id if user entered ID is invalid. setStatus() however should operate as normal (set random id if user entered id in invalid/empty)

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
@@ -184,7 +184,11 @@ class NeuroID private constructor(
         }
     }
 
-    fun setUserID(userId: String, checkForEmpty: Boolean = true): Boolean {
+    fun setUserID(userID: String): Boolean {
+        return setUserID(userID, true)
+    }
+
+    private fun setUserID(userId: String, checkForEmpty: Boolean): Boolean {
         if (checkForEmpty && !validateUserId(userId)) {
             return false
         }
--- a/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
@@ -172,58 +172,70 @@ class NeuroID private constructor(
     }
 
     fun setRegisteredUserID(registeredUserId: String): Boolean {
-        if (!validateUserId(registeredUserId)) {
+        val validID = setGenericUserID(
+            SET_REGISTERED_USER_ID,
+            registeredUserId,
+        )
+
+        if (!validID) {
             return false
         }
-        val result = setGenericUserID(SET_REGISTERED_USER_ID, registeredUserId)
-        return if (result.originCode != NID_ORIGIN_CODE_FAIL) {
-            this.registeredUserID = result.sessionID
-            true
-        } else {
-            false
-        }
+
+        this.registeredUserID = registeredUserId
+        return true;
     }
 
     fun setUserID(userID: String): Boolean {
         return setUserID(userID, true)
     }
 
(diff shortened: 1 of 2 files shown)


NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -181,7 +184,10 @@ class NeuroID private constructor(
         }
     }
 
-    fun setUserID(userId: String): Boolean {
+    fun setUserID(userId: String, checkForEmpty: Boolean = true): Boolean {
```
cometman, line 187: What's the purpose of checkForEmpty?

kevinsites: I think he is using it because in the startSession he doesn't care if its empty but normally we do and will reject if they send and empty string. 

However @kishisaka-nid - I think we should rethink this because we don't want to expose a public API that the user could send in a true and an empty string. We only want a setUserID that takes in a string and that's it. Happy to jump on a zoom and brainstorm if you'd like

kishisaka-nid: setUserID() is used both externally by apps and internally by startSession(). startSession() will set a random id if the userid is empty, setUserID will fail and do nothing if id is empty/invalid. the checkForEmpty flag is needed to distinguish between the two cases.

kishisaka-nid: yeah good point. ill make it separate.

cometman: Yea that was my assumption, I just saw it was on public api which was why I was asking. Could keep state in memory if they passed in a user Id, which is how i implemented in iOS https://github.com/Neuro-ID/neuroid-ios-sdk/blob/main/NeuroID/NeuroIDClass/Extensions/NIDSession.swift#L128

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -181,7 +184,14 @@ class NeuroID private constructor(
...
 
-    fun setUserID(userId: String): Boolean {
+    fun setUserID(userID: String): Boolean {
+        return setUserID(userID, true)
+    }
+
+    private fun setUserID(userId: String, checkForEmpty: Boolean): Boolean {
+        if (checkForEmpty && !validateUserId(userId)) {
+            return false
+        }
```
kevinsites, line 194: If we don't check for empty, won't this skip validation? Meaning for startSession they could pass an invalid ID and we still allow it?

kishisaka-nid: startSession() calls setGenericUserID() that calls getOriginResult() which ultimately calls validateUserID(). validateUserID will ensure that a bad id is caught and in getOriginResult() will ensure that if a bad id is detected, it will be replaced with a random id.

kevinsites: so are we really calling validate user twice?

kishisaka-nid: yeah I think so in case of properly formed ids for setUserID(). in case of badly formed ids, it will fail out once. for startSession() only once. I think I can move the validate to setUserID(String). let me take a look.

kevinsites (approved):

quacke-nid (approved):

priya-xavier (approved):

PR #249, opened 2024-01-23
Eng 6925 make isSimulator check more robust. 

kishisaka-nid:
Add checks to look for various simulators that use x86 hardware and various other types. Tests forthcoming.

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/utils/RootHelper.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/utils/RootHelper.kt
@@ -211,7 +211,8 @@ class RootHelper {
                 || Build.MODEL.contains("Emulator", true)
                 || Build.DEVICE.contains("Emulator", true)
                 || Build.MODEL.contains("Android SDK built for x86")
-                || Build.BOARD == "QC_Reference_Phone" && !Build.MANUFACTURER.equals("Xiaomi")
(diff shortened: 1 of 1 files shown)


kevinsites (dismissed):

NeuroID/src/main/java/com/neuroid/tracker/utils/RootHelper.kt
```diff
@@ -117,7 +117,9 @@ class RootHelper {
 
     private fun isEmulatorFilesPresent(): Boolean {
         emulatorFiles.forEach { path ->
-            return path.exists()
+           if (path.exists()) {
+               return true
+           }
```
kevinsites, line 122: because this is in a foreach, will it return from the function or just the forEach loop?

kishisaka-nid: some kotlin docs for ref: ```Kotlin has three structural jump expressions:
return by default returns from the nearest enclosing function or [anonymous function](https://kotlinlang.org/docs/lambdas.html#anonymous-functions).
break terminates the nearest enclosing loop.
continue proceeds to the next step of the nearest enclosing loop.```

kevinsites (approved):

priya-xavier (approved):

PR #264, opened 2024-02-08
ENG-7125 - Gyro & Accel Cadence

kevinsites:
This will capture the gyro & acceleration data at a 200ms interval as long as
1. the sdk is started and
2. we have the gyroCadence flag set to true

Also includes @kishisaka-nid sensor fix from https://github.com/Neuro-ID/neuroid-android-sdk/pull/259

Note: This PR also refactored the NIDJobServiceManager into a class and a large refactor of UI tests that were dependent on real network traffic to use a mock network approach


quacke-nid:
When a user goes inactive and the sensors are stopped, will this continue sending the last captured value?

Also, I think Kurt's fix should be changed and the NIDSensorHelper internalize the flag on wether or not it is running.

NeuroID/src/main/java/com/neuroid/tracker/utils/NIDTimerActive.kt
```diff
@@ -8,7 +8,10 @@ import com.neuroid.tracker.storage.getDataStoreInstance
 
 object NIDTimerActive {
     private val timer = object: CountDownTimer(30000, 1000) {
+        var isSensorRunning = false
```
quacke-nid, line 11: Remove and track in `NIDSensorHelper`

NeuroID/src/main/java/com/neuroid/tracker/utils/NIDTimerActive.kt
```diff
@@ -8,7 +8,10 @@ import com.neuroid.tracker.storage.getDataStoreInstance
...
+
         override fun onTick(p0: Long) {
+            isSensorRunning = true
```
quacke-nid, line 14: Remove

NeuroID/src/main/java/com/neuroid/tracker/utils/NIDTimerActive.kt
```diff
@@ -17,6 +20,7 @@ object NIDTimerActive {
             val accelData = NIDSensorHelper.getAccelerometerInfo()
 
             NIDSensorHelper.stopSensors()
+            isSensorRunning = false
```
quacke-nid, line 23: Remove

NeuroID/src/main/java/com/neuroid/tracker/utils/NIDTimerActive.kt
```diff
@@ -31,6 +35,10 @@ object NIDTimerActive {
 
     @Synchronized fun restartTimerActive() {
         timer.cancel()
+        if (!timer.isSensorRunning) {
```
quacke-nid, line 38:
```suggestion
        if (!NIDSensorHelper.isSensorRunning) {
```

quacke-nid: Or even just make `NIDSensorHelper.restartSensors()` safe to call when running...

quacke-nid:
@kishisaka-nid Would you mind reviewing my comments on this PR?

kishisaka-nid:
I think this is sound advice. Moving the isSensorsRunning flag to the NIDSensorHelper and have the helper switch the flag (on restart and on stop) will help make it cleaner. Also check to see if sensors are running already in restartSensors(). Then in the timer all that is required restart sensors in the NIDTimerActive.restartTimerActive()

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
```diff
@@ -59,13 +76,39 @@ object NIDJobServiceManager {
...
+            }
+        }
+    }
+
+    private fun createGyroJobServer(): Job {
+        return CoroutineScope(Dispatchers.IO).launch {
+            while (NeuroID.isSDKStarted && NeuroID.captureGyroCadence) {
+                delay(NeuroID.GYRO_SAMPLE_INTERVAL)
+
+                val gyroData = NIDSensorHelper.getGyroscopeInfo()
```
kishisaka-nid, line 98: have to check if sensors are running and restart it not (or have restart check for you). the 30 sec sensor timer might have killed them.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
```diff
@@ -93,6 +93,9 @@ class NIDJobServiceManager(
     private fun createGyroJobServer(): Job {
         return CoroutineScope(Dispatchers.IO).launch {
             while (NeuroID.isSDKStarted && NeuroID.captureGyroCadence) {
+                if(NIDSensorHelper.isSensorActive()){
```
kishisaka-nid, line 96: Can remove the check. restartSensors() already checks if running and returns.

kishisaka-nid:
got some conflict to address.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
```diff
@@ -13,23 +17,31 @@ import kotlinx.coroutines.delay
...
 import kotlinx.coroutines.launch
 import okhttp3.OkHttpClient
+import org.json.JSONArray
+import org.json.JSONObject
 import retrofit2.Retrofit
 import retrofit2.converter.gson.GsonConverterFactory
 import java.util.concurrent.TimeUnit
 
-object NIDJobServiceManager {
+class NIDJobServiceManager(
```
kishisaka-nid, line 26: Suggest reverting the NIDJobServiceManager class changes here and make this a separate ticket and PR.

quacke-nid (dismissed):

NeuroID/src/main/java/com/neuroid/tracker/callbacks/NIDSensorHelper.kt
```diff
@@ -49,21 +55,31 @@ object NIDSensorHelper {
...
     }
 
     fun restartSensors(nSensors: NIDSensors = nidSensors) {
+        if (sensorActive) {
+            return
+        }
+        // need to unregister the listeners before restarting the sensors.
+        stopSensors()
+
         listener = getNIDGenListener(nSensors, firstValuesGyro, firstValuesAccel)
```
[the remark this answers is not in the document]
kevinsites: I agree, I think it might be because they wanted to ensure `listener` was not null when used in the following line to register the listener. Because technically you could call `restartSensors` without calling the `initSensorHelper` function and it would cause an error. I think we can put this in another ticket to refactor into a class with a proper initializer and then remove this

quacke-nid (approved):

NeuroID/src/main/java/com/neuroid/tracker/storage/NIDDataStoreManager.kt
```diff
@@ -140,24 +138,24 @@ internal object NIDDataStoreManagerImp : NIDDataStoreManager {
...
+                    "EVENT: ${event.type} - ${event.tgs} - $contextString"
                 )
 
                 NeuroID.getInstance()?.captureIntegrationHealthEvent(event = event)
             }
 
             when (event.type) {
                 BLUR -> {
-                    NIDJobServiceManager.sendEventsNow(NIDLogWrapper())
+                    NeuroID.getInstance()?.nidJobServiceManager?.sendEventsNow()
```
cometman, line 155: Does the behvaior of sending events in blur and close_session still need to be separated? Also to confirm, we are sending events even if the SDK is explicitly stopped,, but a close_session event happens?

app/build.gradle
```diff
@@ -144,4 +144,7 @@ dependencies {
     implementation "com.squareup.okhttp3:logging-interceptor:5.0.0-alpha.5"
     implementation "com.squareup.okhttp3:okhttp:5.0.0-alpha.5"
 
+    testImplementation 'junit:junit:4.+'
+
+    androidTestImplementation 'com.squareup.okhttp3:mockwebserver:5.0.0-alpha.5'
```
cometman, line 149: Small nitpick, but just noticed this is an alpha version of the library...that the best option?

app/src/androidTest/java/com/sample/neuroid/us/NIDSchema.kt
```diff
@@ -57,22 +59,23 @@ class NIDSchema {
...
-                    assertEquals(json, 200, code)
-                }
-
-                override fun onFailure(code: Int, message: String, isRetry: Boolean) {
-                    assert(false)
-                }
-            }
-        )
+        // Commenting out because tests shouldn't send traffic to prod. Will put back in once dev url is available
+//        val application = ApplicationProvider.getApplicationContext<Application>()
```
cometman, line 63: For iOS I was exposing isDeveopment to tests, and changing to dev url based on that

` public static var isDevelopment: Bool = false
`

     static func getCollectionEndpointURL() -> String {
        if (NeuroID.isDevelopment) {
            return "https://receiver.neuro-dev.com/c"
        }
        return "https://receiver.neuroid.cloud/c"
    }

`

cometman (changes requested):

NeuroID/src/main/java/com/neuroid/tracker/storage/NIDDataStoreManager.kt
kevinsites: Yes we still want to send events immediately on blur and close session, the difference being close session should send events regardless of the SDK still running

app/build.gradle
kevinsites: I had to use that because Kurt used the alpha version for his okhttp retrofit work and the libraries had to match

app/src/androidTest/java/com/sample/neuroid/us/NIDSchema.kt
kevinsites: I think we should discuss how we want to expose that to customers because it was causing confusion for people the difference between our version of development (actual development) and customer development (using their test key)

cometman (approved):

kevinsites:
Merging because the UI tests change which one will pass each time meaning there is something changing in the environment that needs further diagnosis

PR #267, opened 2024-02-10
ENG-7137 add broadcast receiver to listen for network changes

kishisaka-nid:
Add a broadcast receiver to listen for network changes and store these for later retrieval. When network changes occur, fire a NETWORK_STATE event that contains the WIFI state and connected state. Listener is started in NeuroID init() and state is stored in NeuroID.networkInfo. Json payload below. 

```
{"type":"NETWORK_STATE","ts":1707602647761,"iswifi":true,"isconnected":true}
{"type":"NETWORK_STATE","ts":1707602596013,"iswifi":false,"isconnected":false}
```

ACCESS_NETWORK_STATE permission is a normal permission and does not require runtime permission check.


NeuroID/src/main/java/com/neuroid/tracker/models/NIDNetworkInfo.kt
quacke-nid, line 1: Is this necessary?  `NIDNetworkListener` does not need object to create the event, am I missing other references to this data?

kishisaka-nid: The NIDNetworkInfo is stored in the NeuroID class and can be used to check network state before doing something (sendEvent(), saveEvent(), job manager startup, etc). It currently is not being used or necessary. Want me to remove it?

kishisaka-nid: Was thinking about it. I will remove NIDNetworkInfo and only save the isConnected flag in NeuroID.  Don't need to save the isWifi. I don't thing we will act on this information.

priya-xavier (approved):

PR #270, opened 2024-02-14
ENG-7171 - Remove Event Queue Storage from Shared Prefs

kevinsites:
These changes should closely reflect the changes done in 3.0.6 to prevent OOM exceptions with some slight improvements.

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
@@ -99,7 +99,6 @@ class NIDJobServiceManager(
 
                 val gyroData = NIDSensorHelper.getGyroscopeInfo()
                 val accelData = NIDSensorHelper.getAccelerometerInfo()
-                val attrsObj = JSONObject().put("interval", "${1000 * GYRO_SAMPLE_INTERVAL}s")
 
                 dataStore.saveEvent(
                     NIDEventModel(
@@ -107,7 +106,11 @@ class NIDJobServiceManager(
                         ts = System.currentTimeMillis(),
                         gyro = gyroData,
                         accel = accelData,
-                        attrs = JSONArray().put(attrsObj)
+                        attrs = listOf(
+                            mapOf(
+                                "interval" to "${1000 * GYRO_SAMPLE_INTERVAL}s"
+                            )
+                        )
                     )
                 )
 
@@ -179,12 +182,14 @@ class NIDJobServiceManager(
             val event = NIDEventModel(
                 type = LOW_MEMORY,
                 ts = System.currentTimeMillis(),
-                attrs = JSONArray().put(JSONObject().apply {
-                    put("isLowMemory", memoryInfo.lowMemory)
-                    put("total", memoryInfo.totalMem)
-                    put("available", memoryInfo.availMem)
-                    put("threshold", memoryInfo.threshold)
-                })
+                attrs = listOf(
+                    mapOf(
+                        "isLowMemory" to memoryInfo.lowMemory,
+                        "total" to  memoryInfo.totalMem,
+                        "available" to  memoryInfo.availMem,
+                        "threshold" to  memoryInfo.threshold,
+                    )
+                )
             )
 
             return event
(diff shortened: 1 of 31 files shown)


kevinsites:
Tested the GSON library and pushed new commit to fix JSON string being created

OLD Code pre find and replace:
```
        {"siteId":"",
            "clientId":"test",
            "pageTag":"",
            "pageId":"",
            "tabId":"",
            "responseId":"5615139d-59d8-481f-85a6-43c87e7de2db"
            "url":"android://",
            "jsVersion":"5.0.0",
            "sdkVersion":"5.android-null3.1.0",
            "environment":"",
            "jsonEvents":
                ["{\"type\":\"TEST_EVENT\",\"ts\":1}"],
        }
```


GSON without find and replace
```
 {"siteId":"",
            "clientId":"test",
            "pageTag":"",
            "pageId":"",
            "tabId":"",
            "responseId":"c6c967e3-e168-4de3-84ca-3a0250be2a87",
            "url":"android://",
            "jsVersion":"5.0.0",
            "sdkVersion":"5.android-null3.1.0",
            "environment":"",
            "jsonEvents":[
                {"type":"TEST_EVENT","ts":1}
            ]
        }

```

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
```diff
@@ -155,11 +157,55 @@ class NIDJobServiceManager(
...
             }
         }
     }
+
+
+    /**
+     * Get the current system memory state.
+     */
+    private fun checkMemoryLevel(context: Context): NIDEventModel? {
+        val activityManager = context.getSystemService(Context.ACTIVITY_SERVICE) as ActivityManager
```
cometman, line 172: The activityManager var being allocated here is being assigned this system level singelton every time the method is called (using more memory). This should be probably be allocated once at the begining of the file.

cometman (changes requested):

kevinsites:
Regarding Commit: 75c8e51 - Because we were putting objects that were already JSON in our class and then Converting to JSON it was causing some weird issues I.e.
```
{"siteId":"",
            "clientId":"test",
            "pageTag":"",
            "pageId":"",
            "tabId":"",
            "responseId":"01d941a5-fca8-4e7b-a015-38a93adf9020",
            "url":"android://",
            "jsVersion":"5.0.0",
            "sdkVersion":"5.android-null3.1.0",
            "environment":"",
            "jsonEvents":[
                {"type":"TEST_EVENT",
                    "attrs":{
                        "myArrayList":[
                            {"map":
                                {"isLowMemory":true,
                                    "total":1234,
                                    "available":"memoryInfo.availMem",
                                    "threshold":"memoryInfo.threshold"
                                }
                            }
                        ]
                    },
                    "ts":1
                }
            ]
        }

```

Refactored to just use lists/maps so the conversion happens properly
```

        {"siteId":"",
            "clientId":"test",
            "pageTag":"",
            "pageId":"",
            "tabId":"",
            "responseId":"f4ed7877-a6d5-4e8f-bbdb-bdcc56a840dc",
            "url":"android://",
            "jsVersion":"5.0.0",
            "sdkVersion":"5.android-null3.1.0",
            "environment":"",
            "jsonEvents":[
                {"type":"TEST_EVENT",
                    "attrs":[
                        {"isLowMemory":true,
                            "total":"memoryInfo.totalMem",
                            "available":1213,
                            "threshold":45
                        }
                    ],
                    "ts":1
                }
            ]
        }
```

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
kevinsites: Good catch, init with start job but to make doubly sure that we won't break I put in a backup use case. This can be refactored when we turn the jobServiceManager into a proper class with an init

priya-xavier (dismissed):
LGTM!

cometman (dismissed):
Nice cleanup!

kishisaka-nid:
Add this to startSession() to start the job service and sensors else devs that call startSession() only will not send events back. 
```
application?.let {
            nidJobServiceManager.startJob(it, clientKey)
        }
        isSDKStarted = true
```

quacke-nid (dismissed):
LGTM!  Only comments are the `touches` that are still JSON strings and the `var`s for NIDSensorModel are no longer necessary.

quacke-nid (approved):

quacke-nid (dismissed):

kishisaka-nid (approved):
🚢

quacke-nid (approved):

PR #271, opened 2024-02-14
ENG-7267 - Service Tracker Refactor

kevinsites:
1. Moved all attributes from NIDServiceTracker to main NeuroID class. Put them in the companion object so they are static and available through NeuroID.siteID instead of NeuroID.getInstance()?.siteID (all are internal vars)
2. Refactored NIDServiceTracker into the NIDEventSender class and made an interface NIDSenderService to capture what methods are available
3. Updated NIDEventSender to break out API call method (sendEvents) and retryRequest logic for re-use. Additionally Moved the createPayloadJSON method to this class as JobServiceManager was not responsible for creating this just triggering the API call
4. Initialized NIDJobManager with the NIDSenderService object upon NeuroID init
5. Updated tests to all pass and use new mocking of SenderService where applicable

Happy to go over any changes


NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -167,8 +193,14 @@ class NeuroID private constructor(
     }
 
     @VisibleForTesting
-    fun setTestURL(endpoint: String){
-        nidJobServiceManager.endpoint = endpoint
+    fun setTestURL(newEndpoint: String){
+        endpoint = newEndpoint
```
quacke-nid, line 197: Is this assignment necessary?  Can `newEndpoint` be passed directly to the sender?

quacke-nid (dismissed):

kishisaka-nid (commented):
Running a long run test. Ram usage has increased from ~60mb to about ~100mb in an hour. OS doubled the heap to 256mb which is normal.  May want to look at the sensor cadence code later. Other than, looks good. 🚢

kishisaka-nid (approved):

NeuroID/src/main/java/com/neuroid/tracker/models/NIDEventModel.kt
```diff
@@ -162,9 +162,9 @@ data class NIDEventModel(
 }
 
 data class NIDSensorModel(
-    val x: Float?,
-    val y: Float?,
-    val z: Float?
+    var x: Float?,
+    var y: Float?,
+    var z: Float?
```
quacke-nid, line 167: This change is no longer needed, recommend reverting.

quacke-nid (approved):

kevinsites:
The failing test is sporadic and will be fixed in a future PR

quacke-nid (approved):

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -116,6 +120,28 @@ class NeuroID private constructor(
...
+
+
+        @get:Synchronized
+        @set:Synchronized
+        internal var screenFragName = ""
+
+
+        internal var environment = ""
+        internal var siteID = ""
+        internal var rndmId = "mobile"
```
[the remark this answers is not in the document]
kevinsites: I think this was used for the tabID and then DS said tabIDs are useless in mobile. Can move to static in a future PR

kishisaka-nid:
Sensor cadence code is fine. The memory increase was caused by the IntegrationHealthReport which was storing all the events. 🚢

kishisaka-nid (approved):

quacke-nid (approved):

PR #273, opened 2024-02-21
ENG-7251 - RN & Android Consolidation

kevinsites:
We had 89% of the same code in both the RN section and Android sections. The only real difference was checking a few RN components and NOT registering activities in RN to avoid some double registration


NeuroID/src/main/java/com/neuroid/tracker/events/TouchEventManager.kt
```diff
@@ -90,71 +128,198 @@ abstract class TouchEventManager {
         return pointerObj
     }
 
-    private fun generateYValues(motionEvent: MotionEvent): Map<String, Any> {
+    private fun generateXYValues(key: String, motionEvent: MotionEventValues): Map<String, Any> {
```
kishisaka-nid, line 131: This is nice!

NeuroID/src/main/java/com/neuroid/tracker/events/TouchEventManager.kt
```diff
@@ -90,71 +128,198 @@ abstract class TouchEventManager {
...
+                    shouldSaveEvent = true
+                    eventType = TOUCH_START
+                }
+                MotionEvent.ACTION_UP -> {
+                    if (lastTypeOfView > 0) {
+                        lastTypeOfView = 0
+                        lastViewName = ""
+
+                        shouldSaveEvent = true
+                        eventType = TOUCH_START
```
kishisaka-nid, line 224: This should be eventType = TOUCH_END since this is an ACTION_UP.  This will fix the test09ValidateTouchEnd() test.

kishisaka-nid:
Pulled, run, address that comment and we should be good.

NeuroID/src/main/java/com/neuroid/tracker/events/TouchEventManager.kt
kevinsites: Oh good catch, I must have missed that and the TOUCH_MOVE, updated now!

kishisaka-nid (approved):
Tests passing locally, pulled, tested, looks good. 🚢

cometman (approved):
lgtm!

PR #280, opened 2024-03-06
ENG-6428 check for connectivity and pause or resume event collection

kishisaka-nid:
Check for connectivity and pause or resume collection as needed.
Update tests to test this logic.


NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -646,15 +646,15 @@ private constructor(
     }
 
     @Synchronized
-    fun pauseCollection() {
+    fun pauseCollection(flushEvents: Boolean = true) {
```
kevinsites, line 649: Do we want this to be public? The iOS repo made an internal version of this that allows the option of ForceSend or not, while the public method is always true

NeuroID/src/main/java/com/neuroid/tracker/service/NIDNetworkListener.kt
```diff
@@ -18,14 +20,63 @@ import java.util.Calendar
...
+                delay(sleepInterval)
+                neuroID.pauseCollection(false)
+            }
+        } else {
+            if (!neuroID.isStopped()) {
+                return
+            }
+            haveNetworkJob = CoroutineScope(dispatcher).launch {
+                delay(sleepInterval)
+                neuroID.resumeCollection()
```
kevinsites, line 79: I think we need to add another condition here because the SDK could be stopped because they called stopSession and then the network disconnects and reconnects in which case we don't want to start collecting. In iOS we check if there is a connection, the SDK is NOT running, AND there is a userID already set which would indicate our session should still be able to continue

kevinsites (changes requested):

NeuroID/src/main/java/com/neuroid/tracker/service/NIDNetworkListener.kt
kishisaka-nid: Right, good catch! Thanks. Will add that in there.

kishisaka-nid: Ah I see, right. Good point. I'll put in a check to see if the userID is set and if not, simply return.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: can make this happen.

kevinsites (approved):

priya-xavier (approved):

PR #281, opened 2024-03-06
ENG-7408 - Event Send Channel

kevinsites:
Changes include:
1. New Send Event Channel to help reduce multiple request calls (first commit)


NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
```diff
@@ -48,61 +51,110 @@ class NIDJobServiceManager(
...
+    /**
+     * Creates a job that synchronizes all requests to send events.
+     */
+    private fun createSendEventsServer(): Job {
+        return CoroutineScope(dispatcher).launch {
+            try {
+                for (notification in sendEventsNotification) {
+                    var sendNow = notification
+                    var next = sendEventsNotification.tryReceive()
+                    while (next.isSuccess) {
```
kishisaka-nid, line 115: should this be if(next.success())?

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
```diff
@@ -48,61 +51,110 @@ class NIDJobServiceManager(
...
+                        next.getOrNull()?.let {
+                            sendNow = sendNow || it
+                        }
+                        next = sendEventsNotification.tryReceive()
+                    }
```
kishisaka-nid, line 120: Trying to understand why we need the while loop and tryReceive()-ing the next item when we are already iterating in the notification queue for all known values. The next 5 second iteration will pick up the items in the queue after the current iteration. It looks like we want to get all successful entries during the time that we process the items already in the queue?

kishisaka-nid:
do we have to stop the channel when the session is stopped and recreate it when the session is started or can we keep it open?

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
kevinsites: Great question, the idea is that if we have called sendEvents multiple times (lets say 5 times) within the 5 seconds either due to the 5 seconds being hit or the forceSend parameter, we don't actually want to make 5 requests, so instead we do a loop to combine all the requests into one and only make one request to our collection service to be more efficient

kishisaka-nid (dismissed):
LGTM, pulled, tested, working 🚢

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
quacke-nid: Whoever wrote this code should have commented it.  Can you correct that mistake?

kishisaka-nid (approved):
LGTM 🚢

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -841,12 +841,12 @@ private constructor(
...
             BLUR -> {
                 ioDispatcher.launch {
-                    nidJobServiceManager.sendEventsNow()
+                    nidJobServiceManager.sendEvents()
                 }
             }
             CLOSE_SESSION -> {
                 ioDispatcher.launch {
-                    nidJobServiceManager.sendEventsNow(true)
+                    nidJobServiceManager.sendEvents(true)
```
quacke-nid, line 849: Is `ioDispatcher.launch` still needed?

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
```diff
@@ -48,61 +51,112 @@ class NIDJobServiceManager(
...
+        gyroCadenceJob = createGyroCadenceServer()
+    }
+
+    /**
+     * Notify the send events server to initiate a send.
+     */
+    fun sendEvents(forceSendEvents: Boolean = false) {
+        CoroutineScope(dispatcher).launch {
+            sendEventsNotification.send(forceSendEvents)
+        }
```
quacke-nid, line 103: Since `sendEventNotification` has an unlimited buffer, I think we can drop the coroutine and use `trySend`, from [Channel](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-channel/):

`[Sending](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-channel/send.html) to this channel never suspends, and [trySend](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-channel/try-send.html) always succeeds.`

quacke-nid (approved):

kishisaka-nid (approved):

PR #284, opened 2024-03-11
ENG-7027-Include Read Phone state permission in sdk manifest

priya-xavier:
Adding READ_PHONE_STATE permission into SDK manifest for customer awareness.


NeuroID/src/main/AndroidManifest.xml
```diff
@@ -5,4 +5,5 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
+    <uses-permission android:name="android.permission.READ_PHONE_STATE" />
```
kishisaka-nid, line 8: I think this should be only in the app level (sample app). The docs will explain to the user that if they want to pull these signals to add the permission to their app and request permissions from the user. Looks like the sample app already has this in there.

PR #286, opened 2024-03-12
Turn off gyro cadence capture for release

kevinsites:


priya-xavier (approved):

kishisaka-nid (approved):
All that to change one line             `internal val captureGyroCadence = false` approved!

cometman (approved):

PR #287, opened 2024-03-12
ENG-7467 various bug fixes, android

kishisaka-nid:
startSession in advanced library fix to allow null sessionID in startSession

change delay interval from 10s to 2s in network listener 

removed text watch debug` logger.d(msg = "Activity - after text $sequence")`

Don't allow resume to be called if SDK has not been started

change log level to debug for TouchEvent items.


priya-xavier (approved):
startSession(advancedDeviceSignals = true) is working.

kevinsites (approved):

priya-xavier (approved):

PR #291, opened 2024-03-18
Eng 7491 request gps location

kishisaka-nid:
Add a location request listener to that will poll for locations every 10 seconds when the customer moves 10 meters. The coroutine that controls the location collection is tied to the session lifecycle. 

Will setup the coroutine when start, startSession or resumeCollection is called. 

Will start listening for locations when metadata is collected in createMobileMetadata(). 

Will stop listening and cancel the coroutine when the stop, stopSession or pauseCollection is called. 

All the integration tests pass locally and on the build box. 
image


priya-xavier (dismissed):
LGTM! I'm able to see the coordinates on API 30,31 &28 simulators.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -442,7 +448,11 @@ class NeuroID
             }
             dataStore.saveAndClearAllQueuedEvents()
 
-            this.getApplicationContext()?.let { nidCallActivityListener?.setCallActivityListener(it) }
+            this.getApplicationContext()
+                ?.let { nidCallActivityListener?.setCallActivityListener(it) }
+            locationService?.let { ls ->
+                ls.setupLocationCoroutine(getApplicationContext()?.getSystemService(Context.LOCATION_SERVICE) as LocationManager)
+            }
```
kevinsites, line 455: is there a reason you wouldn't have the locationService.let inside the getApplicationContext you already called?

app/src/androidTest/java/com/sample/neuroid/us/MockServerTest.kt
```diff
@@ -0,0 +1,148 @@
...
+                }
+            }
+
+            if (booleanIsFound) {
+                assert(true) {
+                    "Found the event man!"
+                }
+            } else {
+                assert(false) {
+                    "Failed to find event man!"
```
kevinsites, line 143: Should we put the event type that was not found?

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: can do

app/src/androidTest/java/com/sample/neuroid/us/MockServerTest.kt
kishisaka-nid: can do

kevinsites (approved):

priya-xavier (approved):

PR #292, opened 2024-03-27
ENG 7530 remote config

kishisaka-nid:
Corrected branch name just now.
 
Update the app to use the client key: "key_live_IjLua9hEui4X5SlhlqFomaXE" to pull the the remote config from S3. If the app cannot pull the config for any reason, the app will default to the following config: 

```
data class NIDRemoteConfig(
    @SerializedName(value="call_in_progress")
    val callInProgress: Boolean = true,
    @SerializedName(value="event_queue_flush_interval")
    val eventQueueFlushInterval: Long = 10,
    @SerializedName(value="event_queue_flush_size")
    val eventQueueFlushSize: Int = 1999,
    @SerializedName(value="geo_location")
    val geoLocation: Boolean = true,
    @SerializedName(value="gyro_accel_cadence")
    val gyroAccelCadence: Boolean = false,
    @SerializedName(value="gyro_accel_cadence_time")
    val gyroAccelCadenceTime: Long = 200L,
    @SerializedName(value="request_timeout")
    val requestTimeout: Long = 10) {
}

```

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
@@ -199,7 +199,7 @@ class NeuroID
             internal var registeredViews: MutableSet<String> = mutableSetOf()
 
             internal var endpoint = "${Constants.productionEndpoint.displayName}"
-            internal var scriptEndpoint = "${Constants.productionScriptsEndpoint.displayName}"
+            internal var scriptEndpoint = "${Constants.devScriptsEndpoint.displayName}"
             private var singleton: NeuroID? = null
 
             // configuration state
--- a/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
@@ -88,6 +88,21 @@ class NeuroID
         internal var locationService: LocationService? = null
 
         init {
+            // get nid remote config, do this first so we can configure the app
+            NIDConfigurationService(
+                getRetroFitInstance(scriptEndpoint, logger, NIDApiService::class.java),
+                object : OnRemoteConfigReceivedListener {
+                    override fun onRemoteConfigReceived(remoteConfig: NIDRemoteConfig) {
+                        nidSDKConfig = remoteConfig
+                        logger.e("remote_config", "remoteConfig: $remoteConfig")
+                    }
+
+                    override fun onRemoteConfigReceivedFailed(errorMessage: String) {
+                        logger.e("remote_config", "error getting remote config: $errorMessage")
+                    }
+                }, clientKey
+            )
+
             dataStore = NIDDataStoreManagerImp(logger)
             registrationIdentificationHelper = RegistrationIdentificationHelper(this, logger)
             nidActivityCallbacks = ActivityCallbacks(this, logger, registrationIdentificationHelper)
@@ -130,22 +145,11 @@ class NeuroID
             }
 
             // set call activity listener
-            nidCallActivityListener = NIDCallActivityListener(dataStore, VersionChecker())
-            this.getApplicationContext()?.let { nidCallActivityListener?.setCallActivityListener(it) }
-
-            // get nid remote config
-            NIDConfigurationService(
-                getRetroFitInstance(scriptEndpoint, logger, NIDApiService::class.java),
-                object: OnRemoteConfigReceivedListener {
-                    override fun onRemoteConfigReceived(remoteConfig: NIDRemoteConfig) {
-                        nidSDKConfig = remoteConfig
-                        logger.e("remote_config", "remoteConfig: $remoteConfig")
-                    }
-                    override fun onRemoteConfigReceivedFailed(errorMessage: String) {
-                        logger.e("remote_config", "error getting remote config: $errorMessage")
-                    }
-                }, clientKey
-            )
+            if (nidSDKConfig.callInProgress) {
+                nidCallActivityListener = NIDCallActivityListener(dataStore, VersionChecker())
+                this.getApplicationContext()
+                    ?.let { nidCallActivityListener?.setCallActivityListener(it) }
+            }
         }
 
         @Synchronized
@@ -166,9 +170,6 @@ class NeuroID
         companion object {
             var showLogs: Boolean = true
             var isSDKStarted = false
-
-            internal val GYRO_SAMPLE_INTERVAL = 200L
-            internal val captureGyroCadence = false
 
             @get:Synchronized @set:Synchronized
             internal var screenName = ""
@@ -472,8 +473,12 @@ class NeuroID
             dataStore.saveAndClearAllQueuedEvents()
 
             this.getApplicationContext()?.let {
-                nidCallActivityListener?.setCallActivityListener(it)
-                locationService?.setupLocationCoroutine(it.getSystemService(Context.LOCATION_SERVICE) as LocationManager)
+                if (nidSDKConfig.callInProgress) {
+                    nidCallActivityListener?.setCallActivityListener(it)
+                }
+                if (nidSDKConfig.geoLocation) {
+                    locationService?.setupLocationCoroutine(it.getSystemService(Context.LOCATION_SERVICE) as LocationManager)
+                }
             }
             return true
         }
@@ -639,14 +644,19 @@ class NeuroID
 
             dataStore.saveAndClearAllQueuedEvents()
 
-            this.getApplicationContext()?.let { nidCallActivityListener?.setCallActivityListener(it) }
+            if (nidSDKConfig.callInProgress) {
+                this.getApplicationContext()
+                    ?.let { nidCallActivityListener?.setCallActivityListener(it) }
+            }
 
             // we need to set finalSessionID with the set random user id
             // if a sessionID was not passed in
             finalSessionID = getUserID()
 
-            locationService?.let {
-                it.setupLocationCoroutine(getApplicationContext()?.getSystemService(Context.LOCATION_SERVICE) as LocationManager)
+            if (nidSDKConfig.geoLocation) {
+                locationService?.let {
+                    it.setupLocationCoroutine(getApplicationContext()?.getSystemService(Context.LOCATION_SERVICE) as LocationManager)
+                }
             }
 
             return SessionStartResult(true, finalSessionID)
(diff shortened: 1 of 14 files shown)


NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -125,8 +145,11 @@ class NeuroID
             }
 
             // set call activity listener
-            nidCallActivityListener = NIDCallActivityListener(dataStore, VersionChecker())
-            this.getApplicationContext()?.let { nidCallActivityListener?.setCallActivityListener(it) }
+            if (nidSDKConfig.callInProgress) {
+                nidCallActivityListener = NIDCallActivityListener(dataStore, VersionChecker())
+                this.getApplicationContext()
+                    ?.let { nidCallActivityListener?.setCallActivityListener(it) }
+            }
```
kevinsites, line 152: is there a race condition between this and loading the config? Would it be better to turn the config service into an async/await so we can get the config before moving forward?

NeuroID/src/main/java/com/neuroid/tracker/service/NIDConfigurationService.kt
```diff
@@ -0,0 +1,30 @@
...
+    key: String
+): RetrySender() {
+
+    private val responseCallback = object: NIDResponseCallBack {
+        override fun <T> onSuccess(code: Int, response: T) {
+            remoteConfigListener.onRemoteConfigReceived(response as NIDRemoteConfig)
+        }
+
+        override fun onFailure(code: Int, message: String, isRetry: Boolean) {
+            remoteConfigListener.onRemoteConfigReceivedFailed("no can get config for key: $key, use defaults")
```
kevinsites, line 17: Can we change the error message to sound a little more user friendly? Something like "Unable to retrieve the remote configuration for key $key. The default values will be used instead"

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
```diff
@@ -133,7 +132,7 @@ internal class NIDJobServiceManager(
     private fun createSendCadenceServer(): Job {
         return CoroutineScope(dispatcher).launch {
             while (userActive && isActive) {
-                delay(5000L)
+                delay(NeuroID.nidSDKConfig.eventQueueFlushInterval * 1000L)
```
kevinsites, line 135: This would be 10*1000L correct?

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
```diff
@@ -144,15 +143,15 @@ internal class NIDJobServiceManager(
      */
     private fun createGyroCadenceServer(): Job {
         return CoroutineScope(dispatcher).launch {
-            while (NeuroID.isSDKStarted && NeuroID.captureGyroCadence) {
-                delay(NeuroID.GYRO_SAMPLE_INTERVAL)
+            while (NeuroID.isSDKStarted && NeuroID.nidSDKConfig.gyroAccelCadence) {
+                delay(NeuroID.nidSDKConfig.gyroAccelCadenceTime * 1000L)
```
kevinsites, line 147: This would be 200L * 1000L? Isn't that something like 200 seconds?

kevinsites (changes requested):

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -168,9 +188,13 @@ class NeuroID
 
             internal var registeredViews: MutableSet<String> = mutableSetOf()
 
-            internal var endpoint = Constants.productionEndpoint.displayName
+            internal var endpoint = "https://receiver.${Constants.devEndpoint.displayName}"
+            internal var scriptEndpoint = "https://scripts.${Constants.devEndpoint.displayName}"
```
kevinsites, line 192: According to @priya-xavier, the scripts bucket in PROD is `http://scripts.neuro-id.com/`, not the same as collector. Also isn't this setting everything to dev?

kevinsites (changes requested):

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: oh yeah, I forgot to change this back to prod. ill update it. thanks!

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: Good point. I think a blocking approach here might be appropriate to avoid a setting race condition.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDConfigurationService.kt
kishisaka-nid: I agree.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
kishisaka-nid: yeah. ill update the default config to 5.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: separate the script and receiver endpoints in the constants

NeuroID/src/main/java/com/neuroid/tracker/service/NIDJobServiceManager.kt
kishisaka-nid: removed the multiplier. will take the value directly from the config.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -83,6 +90,8 @@ class NeuroID
         internal var locationService: LocationService? = null
 
         init {
+            setRemoteConfig()
```
kevinsites, line 93: I am not a kotlin expert, because this function is marked as run blocking, you don't need an await or anything when you call it do you?

NeuroID/src/main/java/com/neuroid/tracker/utils/Constants.kt
```diff
@@ -9,5 +9,7 @@ enum class Constants(val displayName: String) {
     debugEventTag("Event"),
     fpjsProdDomain("https://advanced.neuro-id.com"),
     productionEndpoint("https://receiver.neuroid.cloud/"),
+    productionScriptsEndpoint("http://scripts.neuro-id.com/"),
     devEndpoint("https://receiver.neuro-dev.com/"),
+    devScriptsEndpoint("http://scripts.neuro-dev.com/")
```
kevinsites, line 14: do we use this anywhere or its just in for dev work/testing?

NeuroID/src/main/java/com/neuroid/tracker/models/NIDRemoteConfig.kt
```diff
@@ -0,0 +1,20 @@
...
+    val eventQueueFlushSize: Int = 1999,
+    @SerializedName(value="geo_location")
+    val geoLocation: Boolean = true,
+    @SerializedName(value="gyro_accel_cadence")
+    val gyroAccelCadence: Boolean = false,
+    @SerializedName(value="gyro_accel_cadence_time")
+    val gyroAccelCadenceTime: Long = 200L,
+    @SerializedName(value="request_timeout")
+    val requestTimeout: Long = 10) {
+}
```
priya-xavier, line 20: sorry for thinking about this only now but should we have all the default times either only in seconds or ms to avoid confusion?

priya-xavier: According to ChatGPT its okay to have these in different units so maybe not being standard is an accepted standard :)

NeuroID/src/main/java/com/neuroid/tracker/utils/Constants.kt
priya-xavier: Yes, lets make sure that we are using the dev endpoint for testing since an PagerDuty alert is triggered if the test config is not present in prod and throws 404's.

NeuroID/src/main/java/com/neuroid/tracker/models/NIDRemoteConfig.kt
kishisaka-nid: As long as I know which units to use and these are not changed, we should be good. Might be good to standardize. Thank for checking though. If we do change the units, please let me know. Also we should finalize any changes before we release thing thing.

NeuroID/src/main/java/com/neuroid/tracker/utils/Constants.kt
kishisaka-nid: only used for dev testing. we should consider leaving this dev for now so we don't set off pager duty until it is ready.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: await is in the remote config call. you can only call await in a suspend function. 

The runBlocking provides the coroutine that we can call the await in.  We then run the network call on a worker thread and suspend the main thread while the worker thread is doing stuff to avoid a warning from the OS to not run long running operations on the Main thread in the logs. Once the worker thread is done, the await() will release the blocking thread to run again. The output will look like this (logging added which is not in the PR): 

```
2024-03-29 10:50:57.836 16275-16275 System.out              com.sample.neuroid.us                I  remote start
2024-03-29 10:50:58.427 16275-16310 ****** Neu...***** init com.sample.neuroid.us                D  remoteConfig: NIDRemoteConfig(callInProgress=true, eventQueueFlushInterval=5, eventQueueFlushSize=2000, geoLocation=true, gyroAccelCadence=false, gyroAccelCadenceTime=200, requestTimeout=10)
2024-03-29 10:50:58.428 16275-16275 System.out              com.sample.neuroid.us                I  remote end
2024-03-29 10:50:58.428 16275-16275 System.out              com.sample.neuroid.us                I  remote store mgr start
2024-03-29 10:50:58.540 16275-16275 System.out              com.sample.neuroid.us                I  remote store mgr end
```

kevinsites: Sweet, as long as we have mitigated that race condition I am good

kevinsites (dismissed):

priya-xavier (approved):

kevinsites (approved):

priya-xavier (approved):

PR #295, opened 2024-04-04
ENG-7530, set environment earlier

kishisaka-nid:
There is a bug in our test code where the setTestURL is not running on initialize when the NeuroID instance is built. This fix ensures that the proper environment is set before the remote config is pulled during unit test run. Setting the config in @Before in the integration test is too late. The remoteConfig is already pulled. The remote config is pulled on init() in NeuroID right after the constructor. 

Our customers will not need to  make changes to their code because the production environment is set as default.

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
@@ -51,8 +51,9 @@ import kotlinx.coroutines.runBlocking
 
 class NeuroID
     private constructor(
-        internal var application: Application?,
+        internal val application: Application?,
         internal var clientKey: String,
+        serverEnvironment: String = "production"
     ) {
         @Volatile internal var pauseCollectionJob: Job? = null // internal only for testing purposes
         private val ioDispatcher: CoroutineScope = CoroutineScope(Dispatchers.IO)
@@ -90,6 +91,17 @@ class NeuroID
         internal var locationService: LocationService? = null
 
         init {
+            when (serverEnvironment) {
+                "production" -> {
+                    endpoint = "${Constants.productionEndpoint.displayName}"
+                    scriptEndpoint = "${Constants.productionScriptsEndpoint.displayName}"
+                }
+                "development" -> {
+                    endpoint = "${Constants.devEndpoint.displayName}"
+                    scriptEndpoint = "${Constants.devScriptsEndpoint.displayName}"
+                }
+            }
+
             setRemoteConfig()
 
             dataStore = NIDDataStoreManagerImp(logger)
@@ -173,8 +185,10 @@ class NeuroID
             }
         }
 
-        data class Builder(var application: Application? = null, var clientKey: String = "") {
-            fun build() = NeuroID(application, clientKey)
+        data class Builder(val application: Application? = null,
+                           val clientKey: String = "",
+                           val serverEnvironment: String= "prod") {
+            fun build() = NeuroID(application, clientKey, serverEnvironment)
         }
 
         companion object {
@@ -199,7 +213,7 @@ class NeuroID
             internal var registeredViews: MutableSet<String> = mutableSetOf()
 
             internal var endpoint = "${Constants.productionEndpoint.displayName}"
-            internal var scriptEndpoint = "${Constants.devScriptsEndpoint.displayName}"
+            internal var scriptEndpoint = "${Constants.productionScriptsEndpoint.displayName}"
             private var singleton: NeuroID? = null
 
             // configuration state
@@ -256,6 +270,7 @@ class NeuroID
         @VisibleForTesting
         fun setTestURL(newEndpoint: String) {
             endpoint = newEndpoint
+            scriptEndpoint = Constants.devScriptsEndpoint.displayName
 
             application?.let {
                 nidJobServiceManager.setTestEventSender(getSendingService(endpoint, logger, it))
--- a/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
@@ -51,7 +51,7 @@ import kotlinx.coroutines.runBlocking
 
 class NeuroID
     private constructor(
-        internal val application: Application?,
+        internal var application: Application?,
         internal var clientKey: String,
         serverEnvironment: String = "production"
     ) {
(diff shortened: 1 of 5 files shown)


cometman (approved):

cometman (changes requested):
Tests failing

kishisaka-nid:
Yeah this is weird. Look like it cannot get the fingerprint lib. 

> Could not resolve all files for configuration ':app:androidAdvancedDeviceLibDebugRuntimeClasspath'.

   > Could not find com.fingerprint.android:pro:2.3.2.
See https://docs.gradle.org/7.5/userguide/command_line_interface.html#sec:command_line_warnings
     Searched in the following locations:
1 actionable task: 1 executed
       - https://dl.google.com/dl/android/maven2/com/fingerprint/android/pro/2.3.2/pro-2.3.2.pom
       - https://repo.maven.apache.org/maven2/com/fingerprint/android/pro/2.3.2/pro-2.3.2.pom
       - https://jcenter.bintray.com/com/fingerprint/android/pro/2.3.2/pro-2.3.2.pom
       - https://maven.fpregistry.io/releases/com/fingerprint/android/pro/2.3.2/pro-2.3.2.pom
       - https://jitpack.io/com/fingerprint/android/pro/2.3.2/pro-2.3.2.pom
     Required by:
         project :app > project :NeuroID

kishisaka-nid:
Tried building another PR ([PR#281](https://github.com/Neuro-ID/neuroid-android-sdk/pull/281))  and these are failing as well in the same place (build All Flavors). Maybe there are connectivity issues with the maven repo that is holding the fingerprint library. Will try again tomorrow.

kishisaka-nid:
[Jitpack](https://jitpack.io/#Neuro-ID/neuroid-android-sdk/-SNAPSHOT) is throwing the same error as well. Tried to build the lib there . Here is the log with the error: 
```
See https://docs.gradle.org/7.5/userguide/command_line_interface.html#sec:command_line_warnings
4 actionable tasks: 1 executed, 3 up-to-date
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8 -Dhttps.protocols=TLSv1.2

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:dataBindingMergeDependencyArtifactsAndroidAdvancedDeviceLibDebug'.
> Could not resolve all files for configuration ':app:androidAdvancedDeviceLibDebugRuntimeClasspath'.
   > Could not find com.fingerprint.android:pro:2.3.2.
     Searched in the following locations:
       - https://dl.google.com/dl/android/maven2/com/fingerprint/android/pro/2.3.2/pro-2.3.2.pom
       - https://repo.maven.apache.org/maven2/com/fingerprint/android/pro/2.3.2/pro-2.3.2.pom
       - https://jcenter.bintray.com/com/fingerprint/android/pro/2.3.2/pro-2.3.2.pom
       - https://maven.fpregistry.io/releases/com/fingerprint/android/pro/2.3.2/pro-2.3.2.pom
       - https://jitpack.io/com/fingerprint/android/pro/2.3.2/pro-2.3.2.pom
     Required by:
         project :app > project :NeuroID

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
```

priya-xavier (dismissed):
LGTM! Was able to see config changes in logs when changes were made to dev mobile config for form_dream102.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -173,8 +185,10 @@ class NeuroID
             }
         }
 
-        data class Builder(var application: Application? = null, var clientKey: String = "") {
-            fun build() = NeuroID(application, clientKey)
+        data class Builder(val application: Application? = null,
+                           val clientKey: String = "",
+                           val serverEnvironment: String= "prod") {
```
kevinsites, line 190: everywhere else we have "production" and not "prod? should we change it here?

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -90,6 +91,17 @@ class NeuroID
...
+            when (serverEnvironment) {
+                "production" -> {
+                    endpoint = "${Constants.productionEndpoint.displayName}"
+                    scriptEndpoint = "${Constants.productionScriptsEndpoint.displayName}"
+                }
+                "development" -> {
+                    endpoint = "${Constants.devEndpoint.displayName}"
+                    scriptEndpoint = "${Constants.devScriptsEndpoint.displayName}"
+                }
+            }
```
kevinsites, line 103: if a customer sends in an invalid value will that cause the endpoints to be wrong/not have a value? or will the defaults be to production regardless of a value being passed in that is valid or not?

kevinsites (changes requested):

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: good point. will try.

kishisaka-nid: it will drop out of the when and use production defaults. that is the reason why when we had "prod" in the sample app, it didn't match any of the clauses, drop out of the when block. since production is set by default, it used default. we will be good.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: yeah this should be "production" will update this.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -53,6 +53,7 @@ class NeuroID
     private constructor(
         internal var application: Application?,
         internal var clientKey: String,
+        serverEnvironment: String = "production"
```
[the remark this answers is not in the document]
kishisaka-nid: no, only used in init(). this is basically a pass through for init(). only init() can see this param.

kevinsites (dismissed):

priya-xavier (approved):

kevinsites (dismissed):

priya-xavier (approved):

kevinsites (approved):

PR #297, opened 2024-04-11
ENG-7738 move channel instance creation before starting service

kishisaka-nid:
Move channel creation before starting the job service. This allows the startService to have an instance of Channel to call methods on. 

java.lang.NullPointerException: Attempt to invoke interface method 'kotlinx.coroutines.channels.ChannelIterator kotlinx.coroutines.channels.Channel.iterator()' on a null object reference


kevinsites (approved):

quacke-nid (approved):

priya-xavier (approved):
LGTM! Can't see the error now.

PR #299, opened 2024-04-16
ENG-7752 ensure location flag turns off location info in metadata

kishisaka-nid:
Ensure that a false location flag turns off location info in metadata.

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/utils/NIDMetaData.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/utils/NIDMetaData.kt
@@ -89,9 +89,10 @@ class NIDMetaData(context: Context, private val locationService: LocationService
         }
     }
 
-    internal fun getLastKnownLocation(context: Context) {
-       locationService?.getLastKnownLocation(context, gpsCoordinates,
-           locationManager=context.getSystemService(Context.LOCATION_SERVICE) as LocationManager)
+    internal fun getLastKnownLocation(context: Context, isLocationAllowed: Boolean) {
+        locationService?.getLastKnownLocation(context, gpsCoordinates,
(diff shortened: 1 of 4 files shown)


NeuroID/src/main/java/com/neuroid/tracker/utils/NIDMetaData.kt
```diff
@@ -89,9 +89,10 @@ class NIDMetaData(context: Context, private val locationService: LocationService
         }
     }
 
-    internal fun getLastKnownLocation(context: Context) {
-       locationService?.getLastKnownLocation(context, gpsCoordinates,
-           locationManager=context.getSystemService(Context.LOCATION_SERVICE) as LocationManager)
+    internal fun getLastKnownLocation(context: Context, isLocationAllowed: Boolean) {
+        locationService?.getLastKnownLocation(context, gpsCoordinates,
+            locationManager=context.getSystemService(Context.LOCATION_SERVICE) as LocationManager,
+            isLocationAllowed=isLocationAllowed)
```
kevinsites, line 95: Is there a reason we would pass the config variable down all the way to the location service instead of just not even calling the location service when the location isn't allowed? That way the service remains free from any config concerns and just reports locations. 
It seems weird that we would call the location service with a boolean saying location is NOT allowed instead of just not calling the service in the first place

priya-xavier (approved):
{"authorizationStatus":"unknown","latitude":-1.0,"longitude":-1.0} when geo_location is false. LGTM!

NeuroID/src/main/java/com/neuroid/tracker/utils/NIDMetaData.kt
kishisaka-nid: It made testing easier to pass in the flag (isolate testing to the location service).

kevinsites (approved):

PR #303, opened 2024-04-19
Eng 7409 neuro id interface refactor

kishisaka-nid:
Return a NeuroIDPublic interface containing public methods instead of a returning the concrete implementation to the host app to use. 

Changes: 

- The NeuroID concrete class will remain the same. This is our concrete implementation wrapper to our SDK functionality. 
- Add an interface called NeuroIDPublic that contains method definitions for all of the public methods. The host app will use this to access SDK functionality. 
- NeuroID.Builder.build() will create an instance of NeuroIDImpl and set it to a singleton. The host app calls this as usual. 
- Keep the NeuroID.getInstance() method which returns a NeuroIDPublic interface. The host app will use this to start, stop sessions ,etc.
- Add a NeuroID.getInternalInstance() method which returns the concrete NeuroID instance. This is used internally within the SDK and is not available externally.  
- Update the extensions (storage and advanced device) to implement the NeuroIDPublic interface. Syntactic sugar as it is called.   

TODO:

- Decide if this is a major or minor release. 
- Consider removing the getInstance() method, have build() return the NeuroIDPublic interface and make the host apps manage the singleton themselves instead of us. 
- Consider removing the getInternalInstance() method, any dependency that internal methods need would use dependency injection (though a framework like dagger/hilt or through parameters passing in methods/constructors) to get dependencies from NeuroIDImpl.  

 
Usage:

Same as before:  

```
NeuroID.Builder(<applciation context>, <client key>).build()
NeuroID.getInstance()?.setRegisteredUserID(<register user id>)
.
<activity start>
NeuroID.getInstance()?.startSession(<session id>, true)
.
<do some activity work>
.
NeuroID.getInstance()?.stopSession(<register user id>)
<activity end>
```
The host  app will be working off of the NeuroIDPublic interface backed by the NeuroID instance and not directly working with the NeuroID instance.

Implications: 

- We can remove all of the internal directives in the NeuroID concrete class. 
- We don't have to worry about marking new internal methods as internal. 
- We will not have to worry about exposing implementation to customers that they are not suppose to see reducing confusion to our customers regarding usage. They will only use what they see in the interface.
- We can include user documentation in the interface and have specific implementation documentation in the implemenation.


NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -188,7 +188,13 @@ class NeuroID
         data class Builder(val application: Application? = null,
                            val clientKey: String = "",
                            val serverEnvironment: String = PRODUCTION) {
-            fun build() = NeuroID(application, clientKey, serverEnvironment)
+            fun build() {
+                if (singleton == null) {
+                    val neuroID = NeuroID(application, clientKey, serverEnvironment)
+                    neuroID.setupCallbacks()
+                    setNeuroIDInstance(neuroID)
```
kevinsites: Does this mean we need to update our documentation so that they don't call `setNeuroIDInstance` right after build anymore?

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
```diff
@@ -0,0 +1,189 @@
...
+
+import android.app.Activity
+import androidx.annotation.VisibleForTesting
+import com.neuroid.tracker.models.SessionStartResult
+
+interface NeuroIDPublic {
+    /**
+     * Deprecated, do not use!
+     */
+    @Deprecated("Replaced with getUserID", ReplaceWith("getUserID()"))
```
kevinsites, line 11: Should this point to `getClientID` instead or `getUserID`?

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
```diff
@@ -0,0 +1,189 @@
...
+    fun getClientId():String
+
+    /**
+     * Return the currently set client id (use this one).
+     */
+    fun getClientID():String
+
+    /**
+     * Set a tag for the widget that is not registered.
```
kevinsites, line 20: I don't think this desc is correct

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * Set a user id.
+     */
+    fun setUserID(userID: String): Boolean
+
+    /**
+     * Return the currently set user id.
+     */
+    fun getUserId(): String
```
kevinsites, line 38: This should be deprecated in favor of `getUserID`

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * Return the currently set screen name for the current page in a session.
+     */
+    fun getScreenName(): String
+
+    /**
+     * Register page targets. This will send events targets to the server for the current page.
+     */
+    fun registerPageTargets(activity: Activity)
```
kevinsites, line 66: Should we note this is only for RN? I don't think an Android user would have reason to call this

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Deprecated, do not use!
+     */
+    @Deprecated("setSiteId is deprecated and no longer required")
+    fun setSiteId(siteId: String)
+
+    /**
+     * Return the current sessionId.
+     */
+    fun getSessionId(): String
```
kevinsites, line 100: Should this be deprecated and `getSessionID` used to keep in line with other function naming?

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Stop the SDK and close the current session. Return true to indicate that
+     * the SDK is stopped.
+     */
+    fun stop(): Boolean
+
+    /**
+     * Stop the SDK and closes the current session.
+     */
+    fun closeSession()
```
kevinsites, line 136: I don't know that we want this exposed externally since they should be calling `stop` or `stopSession`

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * Return the NeuroID SDK version.
+     */
+    fun getSDKVersion(): String
+
+    /**
+     * Clear the currently set user and register user id.
+     */
+    fun clearSessionVariables()
```
kevinsites, line 162: This should not be exposed to customers

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: yes. that is correct.

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
kishisaka-nid: Yeah not correct, updating. Thanks.

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
kishisaka-nid: Missed that one. Thanks.

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
kishisaka-nid: removed.

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
kishisaka-nid: removed

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
kishisaka-nid: Good point. updated

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
kishisaka-nid: updated

kevinsites (approved):

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
```diff
@@ -0,0 +1,190 @@
...
+     * set to the mock collection endpoint. The remote config endpoint will be updated to point to
+     * the dev remote config endpoint.
+     */
+    fun setTestURL(newEndpoint: String)
+
+    /**
+     * Set a dev testing URL for testing.
+     */
+    @VisibleForTesting
+    fun setTestingNeuroIDDevURL()
```
priya-xavier, line 30: Since this is used only for internal testing does this need to be in the public interface? I suppose we could use this for flexibility when we are testing but just wanted to check if that's the reason.

priya-xavier (approved):
Looks great!

NeuroID/src/main/java/com/neuroid/tracker/NeuroIDPublic.kt
kishisaka-nid: I would have liked to have made this internal however the sandbox sample app that is bundled with the SDK uses this as well. so this needs to be publicly accessible. Will look into fixing this in the future (move that test into a unit test).

PR #305, opened 2024-04-24
Fix release variant

kishisaka-nid:
Jitpack build is currently failing. Release branch is using closeSession() which has been moved to a private scope. This should be changed to stopSession().


priya-xavier (approved):

priya-xavier:
Do you know why the local builds on master pass but the jitpack build failed? Wondering what we can do to catch a jitpack build error.

kevinsites (approved):

kishisaka-nid:
@priya-xavier yeah, we don't do a release variant app build in any of our workflows. The integration tests runs an app dev build to run the integration tests. Jitpack builds all variants in app and NeuroID.

PR #307, opened 2024-05-03
Eng 7716 fpjs latency reporting

kishisaka-nid:
Event when FPJS is accessed to get an ID: 
EVENT: ADVANCED_DEVICE_REQUEST - null - rid=1714721739361.e391WU, c=false, l=1540, ct=wifi

Event when ID is pulled from cache
EVENT: ADVANCED_DEVICE_REQUEST - null - rid=1714721739361.e391WU, c=true, l=0, ct=null


NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -151,6 +152,18 @@ class NeuroID
...
+            application?.let {
+                val connectivityManager =
+                    it.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+                val info = connectivityManager.activeNetworkInfo
+                val isConnectingOrConnected = info?.isConnectedOrConnecting()
+                if (isConnectingOrConnected != null) {
+                    isConnected = isConnectingOrConnected
+                }
+                isWifi = info?.type == ConnectivityManager.TYPE_WIFI
+            }
```
kevinsites, line 166: This isn't related to the FPJS work is it?

kevinsites (commented):
For cached ADV events, in the AdvancedDeviceIDManager LN 65 we don't include a 0 for latency, where do we set that?

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: Nope but this was not there originally and the latency need this info otherwise the wifi state is not accurate. I can move this out to a seperate PR, get the network info initialization merged first, before merging this PR.

kishisaka-nid:
NeuroID.captureEvent() line 904. we set l=0 there.

kevinsites (approved):

PR #317, opened 2024-05-10
ENG-6428 update network info on startup

kishisaka-nid:
This was not done in the previous Network service PR (ENG-6428). ENG-7716 (FPJS latency reporting) needs this to send back FPJS latency data (connection type) if the app just started and the network config did not change.

ENG-7716 (FPJS latency reporting) is based on this branch.

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
@@ -7,6 +7,8 @@ import android.content.Context
 import android.content.IntentFilter
 import android.location.LocationManager
 import android.net.ConnectivityManager
+import android.net.NetworkCapabilities
+import android.os.Build
 import android.view.View
 import androidx.annotation.VisibleForTesting
 import com.neuroid.tracker.callbacks.ActivityCallbacks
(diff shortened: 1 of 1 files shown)


NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -151,6 +152,18 @@ class NeuroID
...
+            //get connectivity info on startup
+            application?.let {
+                val connectivityManager =
+                    it.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+                val info = connectivityManager.activeNetworkInfo
+                val isConnectingOrConnected = info?.isConnectedOrConnecting()
+                if (isConnectingOrConnected != null) {
+                    isConnected = isConnectingOrConnected
+                }
+                isWifi = info?.type == ConnectivityManager.TYPE_WIFI
```
kevinsites, line 165: is wifi enough here? do we want to get the specifics or wifi vs. cell vs. ethernet?

kishisaka-nid: yeah I think that would be good. ill change this to string and add the type directly there.

quacke-nid (approved):

kevinsites (approved):

PR #319, opened 2024-05-20
ENG-7869 advanced multiflow relfection

kishisaka-nid:
Move advanced signal startup to the builder(). Uses reflection to achieve this. start() and startSession() still retain the ability to control the advanced signal. Should consider removing these in the future and have the builder control enabling the advanced signal. 

Had to fix AdvancedDeviceIDManagerServiceTest (the tests that  were exercising getRemoteID()). These tests were ending early due to not using the proper dispatcher to dispatch jobs. The IO dispatcher that was used was not waiting for the tests to finish therefore returning non-deterministic results. Changed to use an unconfined dispatcher which is a synchronized dispatcher (blocks until done) to allow the test framework to process the results instead of just prematurely ending. Also fixed the assertions in these tests (some were incorrect). Also the tests change the delay from 5000ms to 10ms allowing the unit tests to finish much quicker. The SDK will use 5000ms as its delay if failures occur in getting the Advanced lib RID.


NeuroID/src/advancedDeviceLib/java/com/neuroid/tracker/extensions/AdvancedDeviceExtension.kt
```diff
@@ -19,20 +19,13 @@ import com.neuroid.tracker.utils.Constants
  * Return false if not started.
  */
 fun NeuroIDPublic.start(advancedDeviceSignals: Boolean): Boolean {
+    NeuroID.getInternalInstance()?.isAdvancedDevice = advancedDeviceSignals
```
kevinsites, line 22: This function should NOT change the global variable, it should only override the global variable

NeuroID/src/advancedDeviceLib/java/com/neuroid/tracker/extensions/AdvancedDeviceExtension.kt
```diff
@@ -47,6 +40,7 @@ fun NeuroIDPublic.startSession(
     sessionID: String? = null,
     advancedDeviceSignals: Boolean
 ): SessionStartResult {
+    NeuroID.getInternalInstance()?.isAdvancedDevice = advancedDeviceSignals
```
kevinsites, line 43: This function should NOT change the global variable, it should only override the global variable

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -91,6 +95,8 @@ class NeuroID
         internal var isConnected = false
         internal var locationService: LocationService? = null
 
+        internal var isAdvancedDevice = false
+
```
kevinsites, line 99: This should be set in the configure/build method

kishisaka-nid: updated. thanks!

NeuroID/src/advancedDeviceLib/java/com/neuroid/tracker/extensions/AdvancedDeviceExtension.kt
kishisaka-nid: Not sure about this one. I think we should update NeuroID.isRN global variable (member) and we should not call checkThenCaptureAdvancedDevice() here in the extension function but only in the NeuroID.startSession() and NeuroID.start(). It is possible that we will request the ADV RID twice if we set the builder isRN to true an call startSession(sessionID, true).

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -58,6 +58,7 @@ class NeuroID
     private constructor(
         internal var application: Application?,
         internal var clientKey: String,
+        internal var isRN: Boolean,
```
kevinsites, line 61: Why is this isRN? I thought we were setting the ADV `isAdvancedDevice` variable? isRN is set in a different method from the react-native sdk

kevinsites: also should these be `val` instead of `var ` because we don't want to change clientKey or `isAdvancedDevice` after its set?

kevinsites (changes requested):

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: fixed. sorry bout that.

NeuroID/src/advancedDeviceLib/java/com/neuroid/tracker/extensions/AdvancedDeviceExtension.kt
kishisaka-nid: not overriding for now. will convene to simplify the flow later.

NeuroID/src/advancedDeviceLib/java/com/neuroid/tracker/extensions/AdvancedDeviceExtension.kt
kishisaka-nid: not changing the global variable for now. will convene to simplify the flow later.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -485,6 +487,53 @@ class NeuroID
             }
         }
 
+        /**
+         * ported from the iOS implementation
+         */
+        override fun startAppFlow(
+            siteID: String,
+            userID: String?,
```
kevinsites, line 495: Have you merged in master because this shouldn't be new should it?

kevinsites: my bad, I was viewing changes since my last review and this showed which is what confused me, when I looked holistically it was fine

kevinsites (changes requested):

NeuroID/src/test/java/com/neuroid/tracker/NeuroIDClassUnitTests.kt
```diff
@@ -777,6 +778,11 @@ open class NeuroIDClassUnitTests {
 
         NeuroID.isSDKStarted = false
         NeuroID.getInternalInstance()?.clientKey = "abcd"
+
+        val logger = mockk<NIDLogWrapper>()
+        every{logger.e(msg = any())} just runs
+        NeuroID.getInternalInstance()?.logger = logger
```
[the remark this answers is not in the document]
kishisaka-nid: yeah I can do that. will make the change.

NeuroID/src/test/java/com/neuroid/tracker/NeuroIDClassUnitTests.kt
```diff
@@ -1123,6 +1137,13 @@ open class NeuroIDClassUnitTests {
     @Test
     fun testSetUserId_not_empty() {
         unsetDefaultMockedLogger()
+        val logger = mockk<NIDLogWrapper>()
+        every {logger.e(tag=any(), msg=any())} just runs
+        every {logger.w(tag=any(), msg=any())} just runs
+        every {logger.d(tag=any(), msg=any())} just runs
+        every {logger.i(tag=any(), msg=any())} just runs
+        NeuroID.getInternalInstance()?.logger = logger
```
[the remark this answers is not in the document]
kishisaka-nid: same here.

kevinsites (approved):

PR #324, opened 2024-05-30
ENG-7830 throttle inital commit.

kishisaka-nid:
Throttle code draft. Please ignore the network security config. This is for proxy man to see the network traffic and was accidentally was pushed. I will remove this.

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
@@ -37,19 +37,16 @@ import com.neuroid.tracker.extensions.captureIntegrationHealthEvent
 import com.neuroid.tracker.extensions.saveIntegrationHealthEvents
 import com.neuroid.tracker.extensions.startIntegrationHealthCheck
 import com.neuroid.tracker.models.NIDEventModel
-import com.neuroid.tracker.models.NIDRemoteConfig
 import com.neuroid.tracker.models.NIDSensorModel
 import com.neuroid.tracker.models.NIDTouchModel
 import com.neuroid.tracker.models.SessionIDOriginResult
 import com.neuroid.tracker.models.SessionStartResult
 import com.neuroid.tracker.service.LocationService
-import com.neuroid.tracker.service.NIDApiService
 import com.neuroid.tracker.service.NIDCallActivityListener
-import com.neuroid.tracker.service.NIDConfigurationService
+import com.neuroid.tracker.service.NIDRemoteConfigService
 import com.neuroid.tracker.service.NIDJobServiceManager
 import com.neuroid.tracker.service.NIDNetworkListener
 import com.neuroid.tracker.service.NIDSamplingService
-import com.neuroid.tracker.service.OnRemoteConfigReceivedListener
 import com.neuroid.tracker.service.getSendingService
 import com.neuroid.tracker.storage.NIDDataStoreManager
 import com.neuroid.tracker.storage.NIDDataStoreManagerImp
@@ -64,14 +61,11 @@ import com.neuroid.tracker.utils.RandomGenerator
 import com.neuroid.tracker.utils.VersionChecker
 import com.neuroid.tracker.utils.generateUniqueHexId
 import com.neuroid.tracker.utils.getGUID
-import com.neuroid.tracker.utils.getRetroFitInstance
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.async
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
 import java.util.Calendar
 
 class NeuroID
@@ -100,6 +94,8 @@ class NeuroID
         internal var debugIntegrationHealthEvents: MutableList<NIDEventModel> =
             mutableListOf<NIDEventModel>()
 
+        internal var nidRemoteConfigService: NIDRemoteConfigService
+
         internal var isRN = false
 
         // Dependency Injections
@@ -116,10 +112,16 @@ class NeuroID
         internal var isConnected = false
         internal var locationService: LocationService? = null
         internal var networkConnectionType = "unknown"
-        internal var nidSamplingService: NIDSamplingService = NIDSamplingService(logger, RandomGenerator())
+        internal var nidSamplingService: NIDSamplingService
         internal var dispatcher: CoroutineDispatcher = Dispatchers.IO
+        internal var randomGenerator = RandomGenerator()
 
         init {
+
+            nidRemoteConfigService = NIDRemoteConfigService(dispatcher, logger, clientKey)
+
+            nidSamplingService = NIDSamplingService(logger, randomGenerator, nidRemoteConfigService)
+
             when (serverEnvironment) {
                 DEVELOPMENT -> {
                     endpoint = Constants.devEndpoint.displayName
@@ -131,9 +133,7 @@ class NeuroID
                 }
             }
 
-            setRemoteConfig()
-
-            dataStore = NIDDataStoreManagerImp(logger)
+            dataStore = NIDDataStoreManagerImp(logger, nidSamplingService, nidRemoteConfigService)
             registrationIdentificationHelper = RegistrationIdentificationHelper(this, logger)
             nidActivityCallbacks = ActivityCallbacks(this, logger, registrationIdentificationHelper)
             application?.let {
@@ -147,8 +147,9 @@ class NeuroID
                     NIDJobServiceManager(
                         this,
                         dataStore,
-                        getSendingService(endpoint, logger, it),
-                        logger,
+                        getSendingService(endpoint, logger, it,
+                            nidRemoteConfigService.getRemoteNIDConfig().requestTimeout),
+                        logger, nidRemoteConfigService
                     )
             }
 
@@ -179,8 +180,10 @@ class NeuroID
                 )
             }
 
+
+
             // set call activity listener
-            if (nidSDKConfig.callInProgress) {
+            if (nidRemoteConfigService.getRemoteNIDConfig().callInProgress) {
                 nidCallActivityListener = NIDCallActivityListener(dataStore, VersionChecker())
                 this.getApplicationContext()
                     ?.let { nidCallActivityListener?.setCallActivityListener(it) }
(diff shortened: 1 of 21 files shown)


NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
```diff
@@ -0,0 +1,101 @@
...
+    private fun updateConfig() {
+        // figure out if we need to update the config, if last update was over 5 minutes ago
+        // get a new config and replace it, this should block till done.
+        val currentTime = Calendar.getInstance().timeInMillis
+        if ((currentTime - lastUpdatedTime) > (UPDATE_FREQUENCY)) {
+            logger.d(msg="refresh required: ${currentTime - lastUpdatedTime}" )
+            NeuroID.getInternalInstance()?.setRemoteConfig()
+            lastUpdatedTime = currentTime
+        }
+    }
```
kevinsites, line 94: As noted
1. this should be in the config service because its the responsibility of the config service
2. We aren't actually checking every 5 min, for v1 we are only retrieving once per app load.

kevinsites (changes requested):

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Return the sample rate, 0 indicates sample everything, 100 indicates sample nothing!
+     */
+    private fun retrieveSampleRate(siteID: String?): Int {
+        // let updateConfig() figure out if we need to update or not
+        logger.d(msg="update config")
+        updateConfig()
+        logger.d(msg="config updated")
+
```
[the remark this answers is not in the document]
kishisaka-nid: ill add a config service. right now NeuroID is grabbing the config.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
```diff
@@ -0,0 +1,101 @@
...
+        // is the site id passed in a linked site id in the options?, if so return the linked site
+        // id sample rate
+        if (isLinkedSiteID(siteID)) {
+            NeuroID.nidSDKConfig.linkedSiteOptions[siteID]?.let {
+                logger.d(msg="retrieveSampleRate site id is in options, ${it.sampleRate}")
+                return it.sampleRate
+            }
```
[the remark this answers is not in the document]
kishisaka-nid: There should no reason why this would not return a mapped value. The `isLinkedSiteID(siteID)` check will return false if there is no mapping in the options hashmap and continue to the next block below eventually reaching the bottom to return the DEFUALT_SAMPLE_RATE.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -510,21 +511,35 @@ class NeuroID
             }
 
             // immediately flush events before anything else
-            CoroutineScope(Dispatchers.IO).launch {
+            CoroutineScope(dispatcher).launch {
                 nidJobServiceManager.sendEvents(true)
                 saveIntegrationHealthEvents()
             }
+
+            linkedSiteID = siteID
```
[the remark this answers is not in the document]
kishisaka-nid: originally this was set early so startSession() / start() would process the linkedSiteID however now that we don't alter the state of the SDK, just stop event collection at capture() without stoping the SDK, this can go back to the original position. start(). startSession() and startAppFlow() don't need any changes.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
```diff
@@ -0,0 +1,101 @@
...
+     */
+    private fun isLinkedSiteID(siteID: String?): Boolean =
+        NeuroID.nidSDKConfig.linkedSiteOptions.contains(siteID)
+
+    /**
+     * is this a parent site id?
+     */
+    private fun isCollectionSiteID(siteID: String?): Boolean =
+        NeuroID.nidSDKConfig.siteID == (siteID ?: false)
+
```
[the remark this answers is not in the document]
kishisaka-nid: yeah it only checks for exact matches with the remote config's siteID. if it is anything but a perfect match, this returns false and the default values (100) will be used. I will change it to check for null or react match and return true otherwise false to let it fall and use the default value.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
kishisaka-nid: add remote config service.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
```diff
@@ -0,0 +1,101 @@
...
+            NeuroID.nidSDKConfig.linkedSiteOptions[siteID]?.let {
+                logger.d(msg="retrieveSampleRate site id is in options, ${it.sampleRate}")
+                return it.sampleRate
+            }
+        }
+        // is the site id passed in a parent site id? if so, return the parent site id sample rate.
+        if (isCollectionSiteID(siteID)) {
+            logger.d(msg="retrieveSampleRate parent site, ${NeuroID.nidSDKConfig.sampleRate}")
+            return NeuroID.nidSDKConfig.sampleRate
+        }
```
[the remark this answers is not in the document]
kishisaka-nid: done.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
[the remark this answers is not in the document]
kishisaka-nid: done.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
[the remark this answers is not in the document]
kishisaka-nid: remote config service added.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -510,21 +511,35 @@ class NeuroID
...
                     }
                 } else {
+                    // we do however have to check sampling status if we are currently
+                    // sampling and we might have to stop it since we are not going through
+                    // startSession() in this block
+                    nidSamplingService.updateIsSampledStatus(siteID)
+                    logger.d(msg="startAppFlow() sdk started ${nidSamplingService.isSessionFlowSampled()}")
+                    if (!nidSamplingService.isSessionFlowSampled()) {
+                        stopSession()
+                        return SessionStartResult(false, userID ?: "")
```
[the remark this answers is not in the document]
kishisaka-nid: updated, isSessionFlowSampled() is checked in the data store manager at the saveEvent() and the queueEvent() methods. if isSessionFlowSampled() is true, events will be saved/queued else return without saving/queueing.

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -530,13 +511,14 @@ class NeuroID
                     SessionStartResult(true, userID ?: "")
                 }
 
+            linkedSiteID = siteID
+            nidSamplingService.updateIsSampledStatus(linkedSiteID)
+            logger.d(msg="startSession() isSessionFlowSampled ${nidSamplingService.isSessionFlowSampled()} ${Companion.siteID}")
+
```
kevinsites, line 517: The sampling status should be updated in 3 places, when start is called, when startSession is called, or LN 509-511 where we are just creating a linkedSiteID session. It should happen there because even if the user just calls start/startSession we might have throttling in place

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
```diff
@@ -0,0 +1,85 @@
...
+            return
+        }
+
+        logger.d(msg="updateIsSampledStatus() we are not sampling")
+        isSessionFlowSampled = false
+    }
+
+
+    /**
+     * Return the sample rate, 0 indicates sample everything, 100 indicates sample nothing!
```
kevinsites, line 38: Isn't this backwards? If we are at 100 we should sample everything and if we are at 0 we sample nothing?

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
```diff
@@ -0,0 +1,85 @@
...
+     */
+    private fun retrieveSampleRate(siteID: String?): Int {
+        // is the site id passed in a linked site id in the options?, if so return the linked site
+        // id sample rate
+        if (isLinkedSiteID(siteID)) {
+            configService.getRemoteNIDConfig().linkedSiteOptions[siteID]?.let {
+                logger.d(msg = "retrieveSampleRate site id is in options, ${it.sampleRate}")
+                return it.sampleRate
```
kevinsites, line 46: What if the siteID is in the linkedSiteOptions but the key `sampleRate` is not?
I.e. 
```
{
  "linkedSiteOptions:{
     "site1": { "otherKey":"foo" }
   }
}
```

will this fall back to the DEFAULT_SAMPLE_RATE or will it return undefined/blow up?

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
```diff
@@ -0,0 +1,85 @@
...
+            }
+        }
+        // is the site id passed in a parent site id? if so, return the parent site id sample rate.
+        if (isCollectionSiteID(siteID)) {
+            logger.d(msg = "retrieveSampleRate parent site, ${configService.getRemoteNIDConfig().sampleRate}")
+            return configService.getRemoteNIDConfig().sampleRate
```
kevinsites, line 52: this value could also not exist, are we defaulting it to 100?

kevinsites (changes requested):

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: fixed. thanks!

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
kishisaka-nid: coment fixed.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
kishisaka-nid: added default checks here 
`return configService.getRemoteNIDConfig().sampleRate ?: DEFAULT_SAMPLE_RATE`

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
kishisaka-nid: added default checks here
`return it.sampleRate ?: DEFAULT_SAMPLE_RATE`

kishisaka-nid: As we define more options here, we will need to add It to the NIDLinkedSiteOption data class.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSamplingService.kt
kishisaka-nid: Yeah if the sample_site param is not add, since it is defined in the NIDRemoteConfig with 100, it will return 100.

PR #330, opened 2024-06-20
ENG-8167 api 21 crash fix

kishisaka-nid:
On API 21 devices, the location service will try to invoke the onStatusChanged() method in the LocationListener. The onStatusChanged() was deprecated in API 29 and not invoked in devices at or above API29 however in testing on an API21 device, it is invoked cause the exception show below: 

`java.lang.AbstractMethodError: abstract method "void android.location.LocationListener.onStatusChanged(java.lang.String, int, android.os.Bundle)"`

To fix this we use the LocationListenerCompat which will avoid the need to implement the onStatusChanged() method for all API version down to 21. The docs specify this below: 
 
https://developer.android.com/reference/android/location/LocationListener

```Note that this method only has a default implementation on Android R and above, and this method must still be overridden in order to run successfully on Android versions below R. LocationListenerCompat from the compat libraries may be used to avoid the need to override for older platforms.```

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/service/LocationService.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/service/LocationService.kt
@@ -2,11 +2,11 @@ package com.neuroid.tracker.service
 
 import android.annotation.SuppressLint
 import android.content.Context
-import android.location.LocationListener
 import android.location.LocationManager
 import android.os.Handler
 import android.os.HandlerThread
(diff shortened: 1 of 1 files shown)


cometman (approved):

priya-xavier (approved):

priya-xavier:
Tested on API 21, 28, 31 and 33. Working as expected.

kevinsites (approved):

PR #332, opened 2024-06-24
ENG-8175 check if started and exit start() if so. fix unit tests

kishisaka-nid:
Check if the SDK is already started(). If so, exit immediately. This will stop the starting of multiple data manager queues that might be occurring right now if start() is called twice in a row without stopping.


kevinsites (approved):

kevinsites (changes requested):
Can we add a test to verify this result?

kishisaka-nid:
yeah good idea. will add a startAlreadyStarted() test.

cometman (dismissed):

kevinsites (approved):

cometman (approved):

PR #333, opened 2024-06-27
Eng 7894 dont call fpjs if session not sampled

kishisaka-nid:
Missed these checks to ensure that we don't capture the advanced device id when we are not sampling events.

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/service/NIDSessionService.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/service/NIDSessionService.kt
@@ -76,7 +76,9 @@ internal class NIDSessionService(
             NIDSingletonIDs.retrieveOrCreateLocalSalt()
 
             neuroID.dataStore.saveAndClearAllQueuedEvents()
-            neuroID.checkThenCaptureAdvancedDevice()
+            if (samplingService.isSessionFlowSampled()) {
+                neuroID.checkThenCaptureAdvancedDevice()
+            }
 
             completion()
         }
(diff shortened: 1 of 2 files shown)


NeuroID/src/main/java/com/neuroid/tracker/service/NIDSessionService.kt
```diff
@@ -76,7 +76,9 @@ internal class NIDSessionService(
             NIDSingletonIDs.retrieveOrCreateLocalSalt()
 
             neuroID.dataStore.saveAndClearAllQueuedEvents()
-            neuroID.checkThenCaptureAdvancedDevice()
+            if (samplingService.isSessionFlowSampled()) {
+                neuroID.checkThenCaptureAdvancedDevice()
```
kevinsites, line 80: why don't we do this check in the `NeuroID.captureAdvancedDevice` method (https://github.com/Neuro-ID/neuroid-android-sdk/blob/f50ee90aaa0d779cdfe706b057c5ca9452729948/NeuroID/src/advancedDeviceLib/java/com/neuroid/tracker/extensions/AdvancedDeviceExtension.kt#L61)? otherwise we will have to put this check everywhere it called including in the ADV extension methods

See iOS Implementation: https://github.com/Neuro-ID/neuroid-ios-sdk/blob/119e39b2d6095fe609a67279868cea31c9fa9811/NeuroID/NeuroIDClass/Extensions/NIDAdvancedDevice.swift#L101

kevinsites (changes requested):

NeuroID/src/main/java/com/neuroid/tracker/service/NIDSessionService.kt
kishisaka-nid: I wanted to add it in there That was my first thought however I couldn't think of a way to test it. I can move it back into checkThenCaptureAdvancedDevice() but have to think about how to test this.

kishisaka-nid: updated.

kevinsites (approved):

priya-xavier (approved):

kevinsites (approved):

PR #334, opened 2024-07-08
ENG-6253 - Add setVariable Method

kevinsites:


NeuroID/src/test/java/com/neuroid/tracker/NeuroIDClassUnitTests.kt
```diff
@@ -689,6 +690,18 @@ open class NeuroIDClassUnitTests {
...
+    @Test
+    fun test_setVariable() {
+        NeuroID._isSDKStarted = true
+        setMockedNIDJobServiceManager(false)
+        setMockedDataStore()
+
+        NeuroID.getInstance()?.setVariable("test", "value")
+
+        assertEquals(1, storedEvents.count())
+        assertEquals(true, storedEvents.firstOrNull()?.type === SET_VARIABLE)
```
cometman, line 702: Very minor, but maybe add a test to check the actual key/value of the set_variable?

cometman (changes requested):

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -666,6 +667,17 @@ class NeuroID
             }
         }
 
+        override fun setVariable(
```
cometman, line 670: In iOS the function name is `setCustomVariable`. We probably want these to be the same.

cometman (changes requested):

NeuroID/src/test/java/com/neuroid/tracker/NeuroIDClassUnitTests.kt
kevinsites: updated

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kevinsites: I thought about doing this but iOS is going to change because we want to be like trackjs and they use `setVariable`

kishisaka-nid (approved):

PR #337, opened 2024-07-15
ENG-8311 - Add Log Messages to SDK

kevinsites:


dbroeckelmanpost (dismissed):

NeuroID/src/main/java/com/neuroid/tracker/service/NIDValidationService.kt
```diff
@@ -45,4 +45,29 @@ internal class NIDValidationService(
 
         return true
     }
+
+    fun scrubIdentifier(identifier: String): String {
+        val emailRegex = Regex("[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}")
+        val ssnRegex = Regex("\\b\\d{3}-\\d{2}-\\d{4}\\b")
+
```
priya-xavier, line 52: should the regex include any 9 digit possible values without the hyphens?

kevinsites: That's a fair question, tbh this was just a quick check for SSN, even if they do set it we would have no way of really knowing that it's an SSN vs. a 9 digit identifier that the customer uses which is why I didn't add more vigorous regex

priya-xavier: sounds good! Looks Good to me!

priya-xavier (dismissed):

kishisaka-nid:
add test for the scrubber. thanks.

kishisaka-nid (approved):

priya-xavier (approved):

NeuroID/src/test/java/com/neuroid/tracker/service/NIDValidationServiceTest.kt
```diff
@@ -100,4 +100,26 @@ class NIDValidationServiceTest {
...
+
+        Assert.assertEquals("t**t@test.com", value)
+    }
+
+    @Test
+    fun test_scrubIdentifier_ssn() {
+        val value = validationService.scrubIdentifier("123-45-6789")
+
+        Assert.assertEquals("***-**-6789", value)
+    }
```
priya-xavier, line 117: Just curious- Should we scrub the last 4 digits of Ssn and leave the rest because that's more unique than the rest?

priya-xavier (approved):

kishisaka-nid (approved):

PR #360, opened 2024-08-15
ENG-8489 - Add New method to track page and elements in one

kevinsites:
Add secondary track elements method, add register target event

I.E.
```
 NeuroID.compose?.trackPageElements(
        pageName = NewApplicationScreens.personal.name,
        elements = arrayOf(
            Pair(first = "First Name", second = firstName),
            Pair(first = "Date of Birth", second = dob),
        )
    )
```


kishisaka-nid (approved):
🚢

PR #362, opened 2024-08-16
ENG-8497 mitigate crash when MutableList.last() is called when empty

kishisaka-nid:
This PR is to mitigate a crash that MGI is currently seeing in our 3.3.2 SDK. it is only happening on the Samsung S22 device in QE testing however that might just be coincidental and could probably happen to all devices. 

What we think is happening is a race condition where the isFullBuffer() method will check if the event list is empty, then the job manager will context switch to clear the event list calling swapEvents() then context switch back to isFullBuffer() which will call eventList.last() to get the last type, see it is empty and throw the exception below that MGI QE is seeing. To mitigate this, we Synchronize the isFullBuffer() method so it will complete execution without being context switched and add a try/catch to catch the exception and handle it if it happens anyway to ensure that we don't fall into this situation. 

Unit tests include to make sure this behavior is occurring. 

```
Caused by java.util.NoSuchElementException: List is empty.
       at kotlin.collections.CollectionsKt___CollectionsKt.last(_Collections.kt:418)
       at com.neuroid.tracker.storage.NIDDataStoreManagerImp.isFullBuffer(NIDDataStoreManager.kt:157)
       at com.neuroid.tracker.NeuroID.captureEvent$NeuroID_androidAdvancedDeviceLibDebug(NeuroID.kt:886)
       at com.neuroid.tracker.NeuroID.captureEvent$NeuroID_androidAdvancedDeviceLibDebug$default(NeuroID.kt:812)
       at com.neuroid.tracker.NeuroID.addLinkedSiteID$NeuroID_androidAdvancedDeviceLibDebug(NeuroID.kt:786)
       at com.neuroid.tracker.service.NIDSessionService$startAppFlow$2.invoke(NIDSessionService.kt:302)
       at com.neuroid.tracker.service.NIDSessionService$startAppFlow$2.invoke(NIDSessionService.kt:294)
       at com.neuroid.tracker.service.NIDSessionService$clearSendOldFlowEvents$2.invokeSuspend(NIDSessionService.kt:411)
```

### DIFF
--- a/NeuroID/build.gradle
+++ b/NeuroID/build.gradle
@@ -114,7 +114,7 @@ dependencies {
 
     implementation "org.jetbrains.kotlin:kotlin-reflect:1.7.10"
 
-    reactNativeImplementation("com.facebook.react:react-native:0.71.0-rc.0")
+    reactNativeImplementation("com.facebook.react:react-android:0.71.0-rc.0")
 
     advancedDeviceLibImplementation("com.fingerprint.android:pro:2.3.2") {
         exclude group: 'org.jetbrains.kotlin', module: 'kotlin-stdlib'
--- a/NeuroID/build.gradle
+++ b/NeuroID/build.gradle
@@ -114,7 +114,7 @@ dependencies {
 
     implementation "org.jetbrains.kotlin:kotlin-reflect:1.7.10"
 
-    reactNativeImplementation("com.facebook.react:react-android:0.71.0-rc.0")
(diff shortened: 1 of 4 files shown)


quacke-nid (dismissed):

cometman (dismissed):

priya-xavier (approved):
Looking great!

cometman (approved):

PR #373, opened 2024-10-23
Eng 8797 fix integration tests

kishisaka-nid:
Rather large update, but not complicated, that simply moves the sample app (android layout sample app) to the main SDK repo to run SDK tests against a mock web server, records the requests and verifies them. The verification consists of counting the number of events and comparing that count with a known event quantity for the login and signup flow in the sample app.  Future verification work will include comparing types of events returned. 

SDK has been updated to target API 34.


testapp/src/androidTest/java/com/neuroid/example/EventRecorder.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * for now we will simply count the number of events and compare
+     * with an expected event count. In the future we should pass in a
+     * JSON expected result and compare with the events received from
+     * the app.
+     */
+    fun verifyEventList(expectedEventCount: Int) {
+        var eventCount = 0
+        println("printEventList --------start----------")
```
kevinsites, line 24: Do we want these print lines in?

testapp/src/androidTest/java/com/neuroid/example/EventRecorder.kt
```diff
@@ -0,0 +1,38 @@
...
+        eventBuffer.forEach { eventModel ->
+            eventModel.jsonEvents.forEach { jsonEventModel ->
+                eventCount ++
+            }
+        }
+        println("printEventList--------end count: expected: $expectedEventCount, actual: $eventCount ----------")
+        // some emulators will returns different events counts, we can approximate this check to
+        // account for the variance here and get a ballpark result here
+        assertTrue("expectedEventCount: ${expectedEventCount} ~ eventCount: ${eventCount} +-20 ",
+            (eventCount >= expectedEventCount - 20 || eventCount <= expectedEventCount + 20)
```
kevinsites, line 34: Is it worth putting the 20 in a variable we could more easily control?

kevinsites (approved):

testapp/src/androidTest/java/com/neuroid/example/EventRecorder.kt
kishisaka-nid: don't need these, ill remove them.

testapp/src/androidTest/java/com/neuroid/example/EventRecorder.kt
kishisaka-nid: Yeah I can put make em constants. thanks.

kishisaka-nid: I think I need the eval to be && instead of || as well.

kishisaka-nid: Running awesome now!

kevinsites (approved):

.github/workflows/github-actions-unit-test.yml
```diff
@@ -67,16 +67,52 @@ jobs:
       - name: Unit Test NeuroId (ReactNative)
         run: ./gradlew :NeuroID:testReactNativeLibDebugUnitTest
 
-        # Execute unit tests
-      - name: Unit Test
+  android_sdk_integration_test_layout_app:
+    runs-on: [macos-12]
```
priya-xavier, line 71: Don't see a reason to use macos. you can change this to ubuntu-latest.

.github/workflows/github-actions-unit-test.yml
```diff
@@ -67,16 +67,52 @@ jobs:
...
+          profile: Nexus 6P
+          cores: 4
+          ram-size: 4096M
+          heap-size: 2048M
+          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+          disable-animations: true
+          script: ./gradlew :testapp:connectedandroidLibDebugAndroidTest
+
+  android_sdk_integration_test_sample_app:
+    runs-on: [ macos-12 ]
```
priya-xavier, line 96: same here

.github/workflows/github-actions-unit-test.yml
kishisaka-nid: According to this article, https://medium.com/@esthcarelle/continuous-delivery-for-android-using-github-actions-effortless-android-testing-building-and-d3759ce66895, the reactivecircus emulator uses hardware acceleration only on MacOS which seems to make the tests more reliable which is why I moved these to separate sections to run the integration tests in MacOS. With ubuntu, this emulator seems to start up more times then not causing the tests to fail out with the "cannot find root window...." error.

.github/workflows/github-actions-unit-test.yml
kishisaka-nid: According to this article, https://medium.com/@esthcarelle/continuous-delivery-for-android-using-github-actions-effortless-android-testing-building-and-d3759ce66895, the reactivecircus emulator uses hardware acceleration only on MacOS which seems to make the tests more reliable which is why I moved these to separate sections to run the integration tests in MacOS. With ubuntu, this emulator seems to start up more times then not causing the tests to fail out with the "cannot find root window...." error.

priya-xavier (approved):
Looks good! Was able to run testapp locally.

PR #384, opened 2025-01-15
ENG-9195 remove integration health report assets and code.

kishisaka-nid:
Easy PR, most of the changes are removing files. 
remove integration health report assets and code. 
small fix to riskyTest1 (turn off animations) to improve reliability.
Drops the size of the library substantially from a 4mb to 300kb.


NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -627,35 +619,32 @@ class NeuroID
         internal fun getFirstTS(): Long = timestamp
 
         override fun setVerifyIntegrationHealth(verify: Boolean) {
-            this.integrationHealthService?.setVerifyIntegrationHealth(verify)
+            // do nothing, removed from debug, kept for backward compatibility
```
kevinsites, line 622: can we add a deprecation method here and/or a console message for customers?

kishisaka-nid: updated. thanks

kevinsites (approved):

priya-xavier (approved):

PR #410, opened 2025-05-21
ENG-10057 add rest of required events that require immediate flush

kishisaka-nid:
add rest of required events that require immediate event buffer flush.


kevinsites (approved):

priya-xavier (approved):

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -958,15 +964,46 @@ class NeuroID
 
             event.log()
 
+            // force event buffer flush when these events are encountered.
             when (type) {
+                FORM_SUBMIT -> {
+                    nidJobServiceManager.sendEvents(true)
+                }
+                PAGE_SUBMIT -> {
```
priya-xavier, line 972: didn't see this event in your ticket but assuming that this is intentional.

kishisaka-nid: Yes, there were additional types that trigger immediate send() calls as well from Rajiv.

PR #412, opened 2025-05-29
ENG-10127 FPJS id cache expiration pull from remote config

kishisaka-nid:
Pull and set FPJS ID cache expiration time from remote config. May need backend change as well to send this data over if not already in the config. 

key is expected to be: advanced_cookie_expires and value is expected to be an int. 24hrs will be set as default

Remote config excerpt with advancedCookieExpiration: 
`NID remoteConfig: NIDRemoteConfig(callInProgress=true, eventQueueFlushInterval=5, eventQueueFlushSize=2000, geoLocation=false, gyroAccelCadence=false, gyroAccelCadenceTime=200, requestTimeout=10, linkedSiteOptions={form_parks912=NIDLinkedSiteOption(sampleRate=50), form_bench881=NIDLinkedSiteOption(sampleRate=0), form_hares612=NIDLinkedSiteOption(sampleRate=100)}, sampleRate=100, siteID=form_skein469, advancedCookieExpiration=24)`

### DIFF
--- a/NeuroID/src/main/java/com/neuroid/tracker/models/NIDRemoteConfig.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/models/NIDRemoteConfig.kt
@@ -24,4 +24,6 @@ data class NIDRemoteConfig(
     val sampleRate: Int = NIDConfigService.DEFAULT_SAMPLE_RATE,
     @SerializedName(value = "site_id")
     val siteID: String = "",
+    @SerializedName(value = "advanced_cookie_expires")
+    val advancedCookieExpiration: Int = 24,
 )
--- a/NeuroID/src/main/java/com/neuroid/tracker/service/AdvancedDeviceIDManagerService.kt
+++ b/NeuroID/src/main/java/com/neuroid/tracker/service/AdvancedDeviceIDManagerService.kt
@@ -42,6 +42,7 @@ internal class AdvancedDeviceIDManager(
     private val linkedSiteID: String,
     // only for testing purposes, need to create in real time to pass NID Key
     private val fpjsClient: FingerprintJS? = null,
+    private val configService: ConfigService
 ) : AdvancedDeviceIDManagerService {
     companion object {
         internal val NID_RID = "NID_RID_KEY"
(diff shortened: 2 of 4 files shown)


NeuroID/src/main/java/com/neuroid/tracker/models/NIDRemoteConfig.kt
```diff
@@ -24,4 +24,6 @@ data class NIDRemoteConfig(
     val sampleRate: Int = NIDConfigService.DEFAULT_SAMPLE_RATE,
     @SerializedName(value = "site_id")
     val siteID: String = "",
+    @SerializedName(value = "advanced_cookie_expires")
+    val advancedCookieExpiration: Int = 24 * 60 * 60 ,
```
kevinsites, line 28: are we defaulting the time to 24 hours? what is DS new requirement and lets put that as the default

NeuroID/src/main/java/com/neuroid/tracker/service/AdvancedDeviceIDManagerService.kt
```diff
@@ -42,6 +42,7 @@ internal class AdvancedDeviceIDManager(
     private val linkedSiteID: String,
     // only for testing purposes, need to create in real time to pass NID Key
     private val fpjsClient: FingerprintJS? = null,
+    private val configService: ConfigService
```
kevinsites, line 45: because the fpjsClient is only for testing, I think the config service should go before it in the parameter list

kevinsites (changes requested):

NeuroID/src/main/java/com/neuroid/tracker/models/NIDRemoteConfig.kt
kishisaka-nid: setting to 12 hrs like web.

NeuroID/src/main/java/com/neuroid/tracker/service/AdvancedDeviceIDManagerService.kt
kishisaka-nid: updated

kevinsites (approved):

priya-xavier (approved):

PR #425, opened 2025-08-06
Eng 10348 sensor listener crasher

kishisaka-nid:
Index out of bounds exceptions were occuring in the onSensorChanged listener. Most likely the Z axis was missing as the index out of bounds was occuring when referencing the 3rd item in the array. These checks should fix these exceptions.


kishisaka-nid:
> Can we add any sort of testing to verify this will not break?

sensor listener tests update with missing x, y and z axis.

PR #472, opened 2026-04-08
ENG-11626 add screen capture service

kishisaka-nid:
Add screen capture detection to the Android SDK. On screen capture detection, a SCREEN_CAPTURE event will be fired. Screen capture support is available for API 34 and above only. Requires android.permission.DETECT_SCREEN_CAPTURE permission. 

Add screen recording detection to the Android SDK. On screen recording detection start, a SCREEN_RECORDING with status=active attribute event will be fired. On screen recording detection stop, a SCREEN_RECORDING with status=inactive attribute event will be fired. Screen recording capture support is available for API 35 and above only. Requires android.permission.DETECT_SCREEN_RECORDING 

details in :https://neuro-id.atlassian.net/browse/ENG-11698


priya-xavier (dismissed):

cpdunphy-nid:
Few questions:
1. Should the SDK declare the permissions in its manifest, or are integrators explicitly expected to add them?
2. May want to review which logs are intended to be preserved as they will be customer facing
3. `contentObserver` and `coroutineScope` still carry dead state
4. Was excluding `VersionChecker.kt` from Sonar coverage intentional long-term?
5. There looks to be a Sonar failure on L92 of `NIDScreenCaptureService.kt‎`

kishisaka-nid:
> Few questions:
> 
> 1. Should the SDK declare the permissions in its manifest, or are integrators explicitly expected to add them?
> 2. May want to review which logs are intended to be preserved as they will be customer facing
> 3. `contentObserver` and `coroutineScope` still carry dead state
> 4. Was excluding `VersionChecker.kt` from Sonar coverage intentional long-term?

1. host apps must declare the permission in the manifest but do not have to include runtime requests in code. 
2. will review logs
3. will remove the states and update test
4. yes since VersionChecked contains system calls that cannot be mocked. that is why the VersionChecker wrapper was created. 
5. will reduce complexity in the method.

jeff-niemann (dismissed):

cpdunphy-nid (approved):

priya-xavier (approved):

PR #473, opened 2026-05-05
ENG-11792 cleanup call in progress event

kishisaka-nid:
Change to cp and events
- removed call in progress ringing events
- change cp = active to cp= connnected
   EVENT: CALL_IN_PROGRESS - null - cp=connected
- change cp = inactive to cp = disconnected
   EVENT: CALL_IN_PROGRESS - null - cp=disconnected


NeuroID/src/main/java/com/neuroid/tracker/service/NIDCallActivityListener.kt
cpdunphy-nid, line 76: Sonar is complaining about the function on L82. Maybe is should be broken up slightly to be more readable?
> Refactor this method to reduce its Cognitive Complexity from 17 to the 15 allowed.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDCallActivityListener.kt
```diff
@@ -75,23 +83,39 @@ class NIDCallActivityListener(
...
             CallInProgress.ACTIVE.state -> {
                 callStateActive = true
-                NIDLog.d(msg = "Call in progress")
-                neuroID.captureEvent(
-                    type = CALL_IN_PROGRESS,
-                    cp = CallInProgress.ACTIVE.event,
-                    attrs = listOf(mapOf("progress" to "active")),
-                )
+                val currentActiveTime = System.currentTimeMillis()
+                if (currentActiveTime - lastActiveTime > 500) {
```
cpdunphy-nid: This is statement will always return true, `lastActiveTime` is never updated so will always resolve to 0.

kishisaka-nid: good catch, the lastInactiveTime reset should only occur inside the if clause.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDCallActivityListener.kt
kishisaka-nid: will break the method down to multiple methods.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDCallActivityListener.kt
```diff
@@ -73,41 +73,26 @@ class NIDCallActivityListener(
...
         when (state) {
-            CallInProgress.INACTIVE.state -> {
+            CallInProgress.DISCONNECTED.state -> {
                 callStateActive = false
                 NIDLog.d(msg = "Call inactive")
                 neuroID.captureEvent(
                     type = CALL_IN_PROGRESS,
-                    cp = CallInProgress.INACTIVE.event,
+                    cp = CallInProgress.DISCONNECTED.event,
                     attrs = listOf(mapOf("progress" to "hangup")),
```
cpdunphy-nid: The `progress` attrs isn't needed anymore, only`direction`

NeuroID/src/main/java/com/neuroid/tracker/events/NIDConstants.kt
cpdunphy-nid: Can we remove the old states? `active`, `ringing`, `inactive`

NeuroID/src/main/java/com/neuroid/tracker/service/NIDCallActivityListener.kt
```diff
@@ -179,7 +164,7 @@ class NIDCallActivityListener(
 
                             // At least one call exists that is dialing, active, or on hold, and no calls are ringing or waiting.
                             TelephonyManager.CALL_STATE_OFFHOOK -> {
-                                saveCallInProgressEvent(CallInProgress.ACTIVE.state)
+                                saveCallInProgressEvent(CallInProgress.CONNECTED.state)
                             }
                             else -> {
                                 saveCallInProgressEvent(CallInProgress.UNKNOWN.state)
```
cpdunphy-nid: If we're already tracking connected and disconnected, do we want to track events that we don't care about?

NeuroID/src/main/java/com/neuroid/tracker/service/NIDCallActivityListener.kt
```diff
@@ -137,15 +122,15 @@ class NIDCallActivityListener(
                 customTelephonyCallback = CustomTelephonyCallback { state ->
                     when (state) {
                         CallInProgress.INACTIVE.state -> {
-                            saveCallInProgressEvent(CallInProgress.INACTIVE.state)
+                            saveCallInProgressEvent(CallInProgress.DISCONNECTED.state)
                         }
 
                         CallInProgress.RINGING.state -> {
                             saveCallInProgressEvent(CallInProgress.RINGING.state)
```
cpdunphy-nid: Don't capture ringing

NeuroID/src/main/java/com/neuroid/tracker/service/NIDCallActivityListener.kt
```diff
@@ -170,7 +155,7 @@ class NIDCallActivityListener(
                     ) {
                         when (state) {
                             TelephonyManager.CALL_STATE_IDLE -> {
-                                saveCallInProgressEvent(CallInProgress.INACTIVE.state)
+                                saveCallInProgressEvent(CallInProgress.DISCONNECTED.state)
                             }
 
                             TelephonyManager.CALL_STATE_RINGING -> {
```
cpdunphy-nid: Don't capture ringing

NeuroID/src/main/java/com/neuroid/tracker/service/NIDCallActivityListener.kt
kishisaka-nid: i can remove progress.

NeuroID/src/main/java/com/neuroid/tracker/events/NIDConstants.kt
kishisaka-nid: think we can. we can use the telephony states and convert for both the newer and older implementations.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDCallActivityListener.kt
kishisaka-nid: will remove

NeuroID/src/main/java/com/neuroid/tracker/service/NIDCallActivityListener.kt
kishisaka-nid: will remove

NeuroID/src/main/java/com/neuroid/tracker/service/NIDCallActivityListener.kt
kishisaka-nid: we can remove the unknown/ring state. if we need to we can add that back in later.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDCallActivityListener.kt
kishisaka-nid: we dont have direction in Android right now. will add in a later PR.

NeuroID/src/main/java/com/neuroid/tracker/service/NIDCallActivityListener.kt
cpdunphy-nid: If `setCallActivityListener` is called a second time while already registered, the permission is granted but `!isReceiverRegistered` is false, so it falls to the else and incorrectly fires an UNAUTHORIZED event

cpdunphy-nid: Dead state

cpdunphy-nid, line 53: What purpose does `NIDPermissionChecker` serve if we are just calling `ActivityCompat` directly?

cpdunphy-nid: Why does this function require `S`, it handles both cases internally?

Something like this may be cleaner:
``` kotlin
private fun registerCustomTelephonyCallback(context: Context) {
    val telephony = context.getSystemService(Context.TELEPHONY_SERVICE) as TelephonyManager
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
        registerWithTelephonyCallback(telephony, context)  // @RequiresApi(S) on this one
    } else {
        registerWithPhoneStateListener(telephony)
    }
}
```

cpdunphy-nid: Checking the version is unnecessary here

kishisaka-nid: good find, we can move the isReceiverRegistered check outside the if statement and exit immediately if already registered.

kishisaka-nid: removed

kishisaka-nid: not used since previous changes, removing.

kishisaka-nid: fixed.

kishisaka-nid: removed due to above fix.

kishisaka-nid: caling the `if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {` in place of the `if (versionChecker.isBuildVersionGreaterThanOrEqualTo31()) {` allows removal of the annotation `@RequiresApi(Build.VERSION_CODES.S)` since the use of Build tells the compiler it is safe since we are checking directly.

cpdunphy-nid (approved):

quacke-nid (approved):

PR #494, opened 2026-08-07
Eng 12112: test enhancements

kurbru-nid:
- Added a Custom Test instrumentation class that enables us to runner deeper test assertions in integration tests
  - This also ensures that the mockwebserver is available immediately upon test app launch. Since our SDK initializes on app creation, setting up the mocks in SETUP is not available in time and we miss visibility into things occurring during init
- Added FPSClient override and extended our Event dispatcher attached to the mock so that it handles all types of requests our sdk is making
- Integrated the setup into the login signup test suite


testapp/src/androidTest/java/com/neuroid/example/EventModel.kt
```diff
@@ -3,6 +3,7 @@ package com.neuroid.example
 data class EventModel(val linkedSiteId: String = "",
                       val siteId: String = "",
                       val clientId: String = "",
+                      val userId: String = "",
```
cpdunphy-nid: `identityId` 😛

kurbru-nid: done

cpdunphy-nid (approved):

kishisaka-nid (approved):

PR #499, opened 2026-08-12
Eng 12101 move events to process observer

kishisaka-nid:
Move the app metadata collection call and the remote configuration capture call into the Process lifecycle observer. This will eliminate session creation on headless app startup and operations (non user startup). Push notifications and device startups may cause app to wake the NeuroID SDK and fire events which we do not want to happen. We only want to see session creation and events when the app is used by a user. 

Adds a log event before capturing noting the call location and `isAdvancedDevice` state. Also addresses a bug where advanced device signals would be captured even if `isAdvancedDevice` is `false.


NeuroID/src/main/java/com/neuroid/tracker/callbacks/ProcessDeviceLifecycleObserver.kt
```diff
@@ -21,5 +21,16 @@ internal class ProcessDeviceLifecycleObserver(
         if (neuroID.isAdvancedDevice) {
             neuroID.checkThenCaptureAdvancedDevice()
         }
+
+        neuroID.captureEvent(type = LOG, m = "isAdvancedDevice setting: ${neuroID.isAdvancedDevice}", level = "INFO")
+
+        neuroID.application?.let { app ->
```
cpdunphy-nid: This can stay in init

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
```diff
@@ -215,15 +215,6 @@ class NeuroID
                         state,
                     )
 
-                metaData =
```
cpdunphy-nid, line 218: Capturing the device metadata (not app metadata), can stay in init

NeuroID/src/main/java/com/neuroid/tracker/callbacks/ProcessDeviceLifecycleObserver.kt
kishisaka-nid: moved back

NeuroID/src/main/java/com/neuroid/tracker/NeuroID.kt
kishisaka-nid: moved back

kurbru-nid (approved):

cpdunphy-nid (approved):

